PR #8, opened 2020-06-12
[YUNIKORN-213] Init source code for yunikorn web-site v2

lamberken:
### Outline

Init source code for yunikorn web-site v2,

Preview
https://apacheyunikorn.github.io/


lamberken:
hi @yangwwei @sunilgovind, this is an inital version, https://apacheyunikorn.github.io/

yangwwei:
hi @lamberken 

I am looking at https://github.com/BigDataArtisans/incubator-yunikorn-site/tree/yk-213, which might be easier. I don't see a README for me to build, and test. WEB-README doesn't seem to. work. Could you please add a README file for the project?

Another thing is, please add Apache License header for all source code files. That's probably the first comment I have : ).

lamber-ken:
> hi @lamberken
> 
> I am looking at https://github.com/BigDataArtisans/incubator-yunikorn-site/tree/yk-213, which might be easier. I don't see a README for me to build, and test. WEB-README doesn't seem to. work. Could you please add a README file for the project?
> 
> Another thing is, please add Apache License header for all source code files. That's probably the first comment I have : ).

Done, thanks

lamberken:
hi @wilfred-s, thanks for your review

> The community, roadmap and download pages are showing up as versioned pages like the docs in this setup which they should not be. 

Right, from docusaurus document, it support create separated community, roadmap, etc. pages.
IMO, it's not a problem to keep them in `docs`.

https://v2.docusaurus.io/docs/creating-pages

> The menu structure seems to be driven of the starting letter of the document. 

As I mentioned, document names are prefixed with letters for categorization. If add a new doc, need update `sidebars.js` also.

> Last point what I do not see yet is the build and publish instructions

After run `yarn build`, the web site will be generated at `build` folder.

wilfred-s:
> Right, from docusaurus document, it support create separated community, roadmap, etc. pages.
> IMO, it's not a problem to keep them in `docs`.
> 
> https://v2.docusaurus.io/docs/creating-pages

The location we store them is not really a problem. The problem is the fact that they are marked as versioned. We cannot mark these files as versioned. We need to do create static files for all the files that are not part of the normal doc revision cycle.
The community, roadmap and download pages are not versioned and there should only ever be one version on the site available. When we create a new revision of the docs we need to exclude them too.

> As I mentioned, document names are prefixed with letters for categorization. If add a new doc, need update `sidebars.js` also.

This needs to be explained and documented somewhere. We should not require a contributor to figure this all out. What we at least need in the documentation is:
- how to name a document
- what tags to add inside the document
- which config to update

> > Last point what I do not see yet is the build and publish instructions
> 
> After run `yarn build`, the web site will be generated at `build` folder.

The question then becomes: does it have all data needed and is that build directory clean enough to commit to the `asf-site` branch? How do we commit that: i.e. do we recursively copy just the `build` dir as the new content of the `asf-site` branch?
We can create a staging setup in the repository which get deployed to an apache staging web server to check if it all works.

lamberken:
hi @wilfred-s,

> The community, roadmap and download pages are not versioned 

I have different opinions on this.

> how to name a document
 
The new site is based on https://v2.docusaurus.io , so developers can read docusaurus directly

> The question then becomes: does it have all data needed and is

You're right, it's ok to copy just the build dir as the new content of the asf-site branch.

lamberken:
hi @wilfred-s, had updated the pr according to your comment, review again, thanks

yangwwei:
I have merged these updates to site-v2 branch.
I played a bit and I have fixed the problem @wilfred-s mentioned, for pages like downloads, roadmap, we do not need a version. I have fixed that on the branch.  That leverages the `blog` plugin, so they are still built via md files, but without versions.
I am doing some more optimizations, but mostly I think it is good. I plan to try to deploy a new version shortly to ASF based on site-v2 branch.

lamber-ken:
> Merged to site-v2 branch. Let's use this branch for some temp time before wiping out the master.
> Thanks a lot @lamberken !

It's ok, welcome. I know `blog` plugin. I think it's ok to let the download, roadmap to link to `next` version.

PR #10, opened 2020-07-28
[YUNIKORN-306] Add yunikorn rest api document

adamantal:
Work done:
- added a new dropdown menu to the site ("API Documentation") 
- added three pages: `Cluster`, `Scheduler` and `System` that list all the available endpoints from `scheduler/routes.go` grouping them by first path item
- used https://github.com/jamescooke/restapidocs for formatting, but I'm open to any suggestions if you have some problems

The examples are extracted via a locally running running a YK scheduler instance, and crawling the 9080 port on my laptop.

### DIFF
--- a/docs/api/cluster.md
+++ b/docs/api/cluster.md
@@ -1,0 +1,62 @@
+---
+id: cluster
+title: Cluster
+---
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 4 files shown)


yangwwei:
hi @adamantal  thanks for the patch, it looks pretty good.
I only have one minor comment about adjusting the sidebar order, but that can be done in a separate JIRA. Thanks!

adamantal:
Let me file a jira for that. What is your suggestion regarding the order?

PR #13, opened 2020-08-05
[YUNIKORN-343] Provide a placement rule that can support attaching NSs to parent queue

adamantal:
Documentation of the `mapping` placement rule from https://github.com/apache/incubator-yunikorn-core/pull/195


wilfred-s:
Closing as the approach is now to use the existing tag rules per YUNIKORN-368

PR #14, opened 2020-08-06
[YUNIKORN-283] Support query params to filter the apps..

manirajv06:
..using queue name/namespace for v1/apps endpoint

### DIFF
--- a/docs/api/scheduler.md
+++ b/docs/api/scheduler.md
@@ -89,7 +89,7 @@ Displays general information about the applications like used resources, queue n
 
 1. queue=<fully qualified queue name\>
 
-Fully qualified queue name filters the applications based on exact match. Passing only leaf queue name doesn't give the expected behaviour. For example, /ws/v1/apps?queue=root.default return the applications running in "root.default" queue.
+The fully qualified queue name used to filter the applications that run within the given queue. For example, "/ws/v1/apps?queue=root.default" returns the applications running in "root.default" queue.
 
 **Auth required** : NO
 
--- a/docs/api/scheduler.md
+++ b/docs/api/scheduler.md
@@ -85,6 +85,12 @@ Displays general information about the applications like used resources, queue n
 
 **Method** : `GET`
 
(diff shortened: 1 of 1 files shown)


docs/api/scheduler.md
```diff
@@ -85,6 +85,12 @@ Displays general information about the applications like used resources, queue n
 
 **Method** : `GET`
 
+**Query Params** : 
+
+1. queue=<fully qualified queue name\>
```
yangwwei, line 90: is \> necessary? or just > is fine here?

docs/api/scheduler.md
```diff
@@ -85,6 +85,12 @@ Displays general information about the applications like used resources, queue n
...
+
+Fully qualified queue name filters the applications based on exact match. Passing only leaf queue name doesn't give the expected behaviour. For example, /ws/v1/apps?queue=root.default return the applications running in "root.default" queue.
```
yangwwei, line 92: > Passing only leaf queue name doesn't give the expected behaviour.

Not completely sure what does this mean. Can you pls refine the text.

docs/api/scheduler.md
manirajv06: > is required to mark the close/end of param value. Backslash is required for escaping to avoid syntax error.

docs/api/scheduler.md
manirajv06: Can we say "Fully qualified queue name filters the applications based on exact match. For example, /ws/v1/apps?queue=root.default return the applications running in "root.default" queue. Passing only leaf queue name ("default") doesn't return the expected results." ?

yangwwei: Ah I see. You want to emphasize the parameter should be using the fully qualified name. How about:

> The fully qualified queue name to filter the applications that run within the given queue. For example, /ws/v1/apps?queue=root.default return the applications running in "root.default" queue.

manirajv06: Looks good.

PR #19, opened 2020-09-03
[YUNIKORN-398] Document namespace based tags

adamantal:
The `namespace.resourcequota` is already documented in `user_guide/resource_quota_mgmt.md`, so I only covered the namespace based annotation in this PR.


wilfred-s:
The tags that are added by the shim should not be part of the rules. The `namespace` tag is used as an example but we have not defined it anywhere, which is a problem.
If we use a different shim than the tag will not be added. We need to find a way to document this as part of shim specific behaviour. Not sure how how we do that but can you come up with a way that we can do this properly that can handle differences between shims etc?

yangwwei:
hi @adamantal 

My suggestion is to add a tutorial for this something like http://yunikorn.apache.org/docs/next/user_guide/resource_quota_management#option-2-11-mapping-from-namespaces-to-queues. So user could follow and configure with this certain placement rule in order to set up `static parent queues + dynamical leaf queue` scenario. As @wilfred-s mentioned, this doesn't need to be in the placement rule doc.

adamantal:
Thanks for the review @wilfred-s and @yangwwei.

I moved this part to a shim specific section, as Wilfred suggested. In this PR I only moved the addition in this patch, but we can also move some other docs as well.
I also modified a bit and added more complex example just like the one in the resource quota management section.

yangwwei:
hi @adamantal 

I still think this should go to the doc: http://yunikorn.apache.org/docs/next/user_guide/resource_quota_management
Add another section just like the `Option 1` and `Option 2`.
This is a user-facing doc, a tutorial. So end-users can simply follow this to set up the queue structure with placement rule easily.
Does that make sense to you?

For the doc itself, can you please include a sample configuration, where we pre-setup some queues as parents, and leverage the placement rule to do auto-placement. Something like

```
placementrules:
  - xxxx
queues:
   root
    - finance
    - marketing
    - dev
```

then tell users what kind of annotations need to be added in their pods. in order to properly place apps to specified queue, e.g "root.finace.{namespace-xyz}"

adamantal:
Moved back the doc to `resource_quota_mgmt.md`. I also modified a bit to be more an example (and renamed to `Option 3`).

Could you please review it?

docs/user_guide/resource_quota_mgmt.md
```diff
@@ -150,3 +150,38 @@ Jobs continue to be submitted to namespaces, based on the `Placementrule` used
 in the configuration. YuniKorn will automatically run the job and all its pods in
 the corresponding queue. For example, if a job is submitted to namespace `development`,
 then you will see the job is running in `root.development` queue.
+
+## Option 3) Custom mapping of workspaces to queues using Kubernetes shim
```
yangwwei, line 154: I do not think we should use `workspace` here, this is not a native concept in K8s, people would get confused about this. Can we replace this with `Hierarchy queues with dynamical leaves that mapped to namespaces`

yangwwei: Can we add the `Goal` and `Configuration` section as well?
In order to keep the format of the doc consistent. I suggest to have a structure like the following:

- Goal: what can be achieved via this setup? From the user's perspective
- Configuration: the example configuration in the current doc
- Set up namespaces: explain how to set up namespace to incorporate this config, e.g adding annotations.
- Run workloads: how to submit jobs to the structure just defined

so the user can simply following step by step to set this up.

docs/user_guide/resource_quota_mgmt.md
```diff
@@ -150,3 +150,38 @@ Jobs continue to be submitted to namespaces, based on the `Placementrule` used
...
+ - name: root
+   queues:
+     - name: namespaces
+```
+and the `production` namespace is annotated with the following annotation:
+```yaml
+yunikorn.apache.org/parentqueue: root.namespaces
+```
+
+Admins can put each application into different user's queue based on the namespace setting using the settings above.
```
yangwwei, line 185: we can just ask users to submit their jobs to namespaces, this is not something required to be done by admins.

docs/user_guide/resource_quota_mgmt.md
```diff
@@ -150,3 +150,38 @@ Jobs continue to be submitted to namespaces, based on the `Placementrule` used
...
+
+Let's say user `bob` submitting an application to the `production` namespace. The application will be placed onto `root.namespaces.production.bob`.
```
yangwwei, line 187: is this correct? I was under the impression the app will be placed to `root.namespaces.production`. Why there is a user queue under production? Can we simplify the example? Change to the path like "root.production.bob".

yangwwei (changes requested):

docs/user_guide/resource_quota_mgmt.md
adamantal: Actually you don't need to annotate the namespace if you want to achieve that.
By changing the queue hierarchy to produce a config which produces the `root.production.bob` as an answer, you don't use the annotation on the namespace at all.

I simplified the example a bit, but we do need that layer - that is what this whole example is about.

docs/user_guide/resource_quota_mgmt.md
adamantal: Removed the admin and the cluster administrator parts.

docs/user_guide/resource_quota_mgmt.md
adamantal: This is not very applicable here, as it makes more sense to first define the namespace config and only after the placement rules (the order is inverted following this structure), but I've modified it as you explained.

adamantal:
Updated the PR from @yangwwei's review items.
Would you be so kind as to review this again? (I hope I resolved your items correctly.)

docs/user_guide/resource_quota_mgmt.md
yangwwei: I mean can we set up the static queues like:

```
placementrules:
 - name: tag
   value: namespace
   create: true
   parent:
     - name: tag
       value: namespace.parentqueue
queues:
 - name: root
   queues:
     - name: production
     - name: development
```

and then we submit an app to a namespace `bob`, then we expect to see the app will be running under `root.production.bob`. Having `namespaces` as the parent queue name is a bit confusing to me.

adamantal: The `bob` namespace is a bit odd, since it's rather a user. I replaced it with `finance` and modified the example. Could you double check @yangwwei ?

adamantal:
Modified the example, could you please review this again @yangwwei ?

docs/user_guide/resource_quota_mgmt.md
yangwwei: Sure. Looks good now. Thanks for making these changes. +1

yangwwei (approved):

PR #20, opened 2020-09-08
[YUNIKORN-399] cleanup website build

wilfred-s:
Upgrade to docusaurus 2.0.0-alpha.63, fix to that version.
Cleanup package.json, remove unused dependencies.

Remove yarn.lock file and rely on published docusaurus dependencies.
Docusaurus pins internal dependencies from alpha.62 onwards. Using a
local yarn.lock could interfere and break that.


wilfred-s:
Changes committed to master

PR #23, opened 2020-09-22
[YUNIKORN-424] Fix nonstandard title grammar for the markdown file

yanghua:
The current grammar will cause the title rendering to fail, the case is here: http://yunikorn.apache.org/docs/next/design/scheduler_core_design/

### DIFF
--- a/docs/design/scheduler_core_design.md
+++ b/docs/design/scheduler_core_design.md
@@ -62,7 +62,7 @@ Scheduler core is agnostic about underneath platforms, all the communications ar
                 +---------------------------------------------+
 ```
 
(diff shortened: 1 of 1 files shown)


yanghua:
@yangwwei

yangwwei (approved):
LGTM. Thanks @yanghua

yangwwei:
hi @yanghua I have created https://issues.apache.org/jira/browse/YUNIKORN-424 and assigned that to you. If the user ID was not correct, pls let me know. We need an associated JIRA for every PR for tracking purposes. Thanks

PR #29, opened 2020-10-13
[YUNIKORN-405] Document config update API

kingamarton:


docs/api/scheduler.md
```diff
@@ -434,6 +434,76 @@ Reponse
...
+        create: true
+    queues:
+      - name: root
+        submitacl: '*'
+        properties:
+          application.sort.policy: stateaware
+checksum: [173,7,83,162,71,12,24,54,61,72,195,227,207,95,85,14,128,101,107,7,112,214,125,208,136,131,212,118,44,191,90,222]
+```
+**Note:** the checksum is the base configuration checksum on top of what we want to make the changes.
+If the provisioned checksum differs from the actual scheduler checksum, the configuration update will fail.
```
yangwwei, line 469: > If the provisioned checksum differs from the actual scheduler checksum, the configuration update will fail.

if the provided checksum in the request differs from the base version, the configuration update will fail. This is to avoid applying any unintentional changes to the running version. Subsequently, the client needs to retrieve the latest version from the scheduler and send another update with the up-to-date checksum.

docs/api/scheduler.md
```diff
@@ -434,6 +434,76 @@ Reponse
...
+
+### Failure response
+
+The configuration update can fail due to different reasons such as: 
+- invalid configuration, 
+- outdated base checksum, 
+- failure while saving the configuration to the scheduler.
```
yangwwei, line 476: this line is quite confusing... I think this is to outline what could be the failures
this will be hard for users to understand why the saving fails..

yangwwei (changes requested):
Hi @kingamarton 

I found we do not have the doc for this API as well:

```
	// endpoint to retrieve the current conf
	route{
		"Scheduler",
		"GET",
		"/ws/v1/config",
		getClusterConfig,
	},
```

could you please add one for it as well?

And another follow up is to change `/ws/v1/validate-conf` to be part of `/ws/v1/config` too, correct? That can be done in a follow up JIRA as that needs both changes in the core and the shim (admission-controller calls it)

kingamarton:
> endpoint to retrieve the current conf

I will add some lines for getClusterConfig.
Related the refactoring of the validate-conf, that will be refactored only during the whole API refactoring, not? Does it make sense to do any kind of smaller refactoring before that?

docs/api/scheduler.md
kingamarton: While saving the scheduler configuration it may fail due to some unexpected errors while updating the partitions, since if the validation passed, it should be OK in most of the cases, but in any case the response message will have the exact error message.
@yangwwei do you want me to remove this line, or how can it be formulated in a more explicit way?

docs/api/scheduler.md
kingamarton: changed it

docs/api/scheduler.md
```diff
@@ -433,6 +433,141 @@ Reponse
...
+- 96
+- 239
+- 55
+- 254
+- 217
+- 41
+- 151
+- 145
+
+sha256 checksum: BAB3D76402827EABE62FA7E4C6BCF4D8DD9552834561B6B660EF37FED9299791
```
yangwwei, line 497: is this correct? Why there is a multi-line checksum?

docs/api/scheduler.md
yangwwei: When it fails to save the configs, that means something is invalid in the config, so it falls to the `invalid configuration`. I guess we can just remove this line, that might be less confusing if I am a user.

yangwwei (changes requested):

docs/api/scheduler.md
kingamarton: removed

docs/api/scheduler.md
kingamarton: This is how the response looks like for getConfig()

yangwwei: Oh, no. I don't think this is correct.
@wilfred-s , could you please take a look at this?

wilfred-s: That is the way the json marshaller does the `[]byte` marshalling. It converts every value in the array to a 8bit integer (0-255) and prints the value on a line. Unmarshalling probably gives you the right thing back again but it is not readable.
In the `getClusterConfig()` call I thus opted for a printed format using the standard base 16, upper-case, two characters per byte conversion.
The config also has ``` Checksum   [32]byte `yaml:"-" json:"-"` ``` added so the marshaller leaves it alone. The yaml version is slightly different than what I added but still human readable. We can use the `TestGetConfigYAML` and `TestGetConfigJSON` to see the impact of the marshaller

kingamarton: Thank you @wilfred-s. I will need to find an another solution for the checksum marshalling problem for the validation, because this is not a user friendly.

yangwwei (approved):
+1

PR #37, opened 2021-02-02
[YUNIKORN-353] Disable version alerts

wilfred-s:
Disable docusaurus version alert for older versions. This is a quick
remove for all versions. Further customisation is possible as per
docusaurus issue #3013.


yangwwei (approved):
+1

PR #38, opened 2021-02-04
[YUNIKORN-524] use apache approved action for push

wilfred-s:
Apache limits the actiona that are allowed in gthub actions. The action
used for the auto push is not on that list. Replace with standard push.


wilfred-s:
The commit message is also updated to show the commit from master that is pushed.

yangwwei (approved):
Thanks for fixing this @wilfred-s .

PR #39, opened 2021-02-11
[YUNIKORN-309] behaviour with two quota systems

wilfred-s:
YuniKorn and Kubernetes Namespace quotas can be configureed at the same
time. Preference to only use YuniKorn quota system.
Explain behaviour when both systems are configured.
Specify which resources are considered during quota enforcement.


ayubpathan:
Looks good to me. Thanks @wilfred-s.

wilfred-s:
committing based on review

PR #41, opened 2021-04-03
[YUNIKORN-618] Add Gang scheduling user and troubleshooting doc

yangwwei:
Some screenshots about the doc pages, tested locally

Gang_Scheduling___Apache_YuniKorn__Incubating_
Gang_Scheduling___Apache_YuniKorn__Incubating__and_wyang____workspace_github_wyang__-_____incubator-yunikorn-site_docs_user_guide_gang_scheduling_md
Gang_Scheduling___Apache_YuniKorn__Incubating_


docs/user_guide/gang_scheduling.md
```diff
@@ -0,0 +1,205 @@
...
+the app’s minimal resource request can be satisfied. Otherwise, apps
+will be waiting in the queue. Apps are queued in hierarchy queues,
+with gang scheduling enabled, each resource queue is assigned with the
+maximum number of applications running concurrently with min resource guaranteed.
+
+![Gang Scheduling](./../assets/gang_scheduling_iintro.png)
+
+:::note
+gang scheduling feature is only available in *0.10.0* and later releases.
+:::
```
kingamarton, line 37: Since we have our documentation versioned, I think we don't need this note.

docs/user_guide/gang_scheduling.md
```diff
@@ -0,0 +1,205 @@
...
+A task group is a “gang” of tasks in an app, these tasks are having the same resource profile
+and the same placement constraints. They are considered as homogeneous requests that can be
+treated as the same kind in the scheduler.
+:::
+
+### Prerequisite
+
+It is recommended to enable gang scheduling at least in the *queue level*, which means for
+each resource queue, either all the apps are gang scheduled, or non of the apps is gang scheduled.
+For the queues which have gang scheduling enabled, the queue sorting policy needs to be set either
```
kingamarton, line 55: `For the queues which have gang scheduling enabled` I think this part is not so clear. It makes me to think that we can enable somewhere the gang scheduling for the queue.

docs/user_guide/gang_scheduling.md
```diff
@@ -0,0 +1,205 @@
...
+On Kubernetes, YuniKorn discovers apps by loading metadata from individual pod, the first pod of the app
+is required to enclosed with a full copy of app metadata. If the app doesn’t have any notion about the first or second pod,
+then all pods are required to carry the same taskGroups info. Gang scheduling requires taskGroups definition,
+which can be specified via pod annotations. The required fields are:
+
+| Annotation                                     | Value |
+|----------------------------------------------- |---------------------	|
+| yunikorn.apache.org/task-group-name 	         | Task group name, it must be unique within the application |
+| yunikorn.apache.org/task-groups                | A list of task groups, each item contains all the info defined for the certain task group |
+| yunikorn.apache.org/schedulingPolicyParameters | Optional. A arbitrary key value pairs to define scheduling policy parameters. Currently only *placeholderTimeoutInSeconds* is supported. It defines the reservation timeout for how long the scheduler should wait until giving up allocating all the placeholders. |
```
kingamarton, line 75: I think here we need to add some clarification related the timeout. It will not define the reservation timeout, since the timer is not started when the app enters the reserving state, but when we have the first placeholder allocation. Also we should mention it's default value.

docs/user_guide/trouble_shooting.md
```diff
@@ -144,6 +144,43 @@ kubectl scale deployment yunikorn-scheduler -n yunikorn --replicas=0
 kubectl scale deployment yunikorn-scheduler -n yunikorn --replicas=1
 ```
 
+## Gang Scheduling
+
+### 1. No placeholders created, app's pods are pending
+
+*Reason*: This is usually because the taskGroups definition is invalid, and the app is rejected by the scheduler.
```
kingamarton, line 151: It can be because the total resource request exceeds the queue resources. In this case the app is also rejected

kingamarton (commented):
Since for the 0.10 release we have a couple of known issues related this feature, I think we should mention them in the documentation as well.

docs/user_guide/trouble_shooting.md
yangwwei: Good point, added to the doc

docs/user_guide/gang_scheduling.md
yangwwei: Make sense, adding more text to explain this.

docs/user_guide/gang_scheduling.md
yangwwei: Removed.

yangwwei:
@kingamarton  doc updated, pls help to review.

docs/user_guide/gang_scheduling.md
yangwwei: Removed that to avoid confusion

kingamarton (approved):

PR #43, opened 2021-04-10
[YUNIKORN-635] Release notes is missing in 0.9.0 Get Started doc.

yangwwei:
Release_Notes___Apache_YuniKorn__Incubating_


wilfred-s:
This is a copy of the already existing page under the [static pages](http://yunikorn.apache.org/release-announce/0.9.0/)
This seems a duplication, can we not point to the same location for that page?

yangwwei:
> This is a copy of the already existing page under the [static pages](http://yunikorn.apache.org/release-announce/0.9.0/)
> This seems a duplication, can we not point to the same location for that page?

That would be the best, but I don't know how to achieve that, unfortunately.

wilfred-s:
In the side bar we can use a normal link like this:
```
        {
          "type": "link",
          "label": "Release Note",
          "href": "/release-announce/0.9.0/"
        },
```
That will open up the existing release note in as a normal page.

yangwwei:
> In the side bar we can use a normal link like this:
> 
> ```
>         {
>           "type": "link",
>           "label": "Release Note",
>           "href": "/release-announce/0.9.0/"
>         },
> ```
> 
> That will open up the existing release note in as a normal page.

Yes, thanks for the pointer. This works nicely.
I've done this for both 0.9.0 and 0.8.0.
Thanks!

wilfred-s (approved):
LGTM
committing this change

PR #47, opened 2021-04-13
[YUNIKORN-405] Extend documentation with checksum

kingamarton:


docs/api/scheduler.md
```diff
@@ -528,13 +528,20 @@ partitions:
         submitacl: '*'
         properties:
           application.sort.policy: stateaware
+checksum: BAB3D76402827EABE62FA7E4C6BCF4D8DD9552834561B6B660EF37FED9299791
 ```
+**Note:** the checksum is the base configuration checksum on top of what we want to make the changes.
+If the provided checksum in the request differs from the base version, the configuration update will fail.
```
wilfred-s, line 534: What is the base configuration? We need to be far clearer in what that is:
```
Updates must use a current running configuration as the base. The base configuration is the configuration version that was retrieved earlier via a GET request and updated by the user. The update request must contain the checksum of the _base_ configuration. If the checksum provided in the update request differs from the currently running configuration checksum the update will be rejected.
```

docs/api/scheduler.md
```diff
@@ -528,13 +528,20 @@ partitions:
...
+This is to avoid applying any unintentional changes to the running version.
+Subsequently, the client needs to retrieve the latest version from the scheduler and send another update with the up-to-date checksum.
 
 ### Failure response
 
-The configuration update can fail due to invalid configuration 
+The configuration update can fail due to different reasons such as:
+- invalid configuration,
+- outdated base checksum.
```
wilfred-s, line 542: It is not an outdated checksum it is the incorrect checksum. We have no idea if it is outdated or just some random string.

docs/api/scheduler.md
```diff
@@ -528,13 +528,20 @@ partitions:
...
 
-In this case the transaction will be rolled back, and the proper 
+In each case the transaction will be rolled back, and the proper
```
wilfred-s, line 544: We do not roll back, the change is never applied. Better to say `rejected`

wilfred-s (changes requested):
text changes to make it a bit clearer

wilfred-s (approved):
LGTM

PR #48, opened 2021-04-13
[YUNIKORN-566] Gang design document

wilfred-s:
PDF version as attached to the jira converted into md file with minor
grammar (commas and dashes) and one typo fix.


docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+| ----------- | --------------------- | ---------------------------------------------- |
+| 2020-11-25  | Wilfred Spiegelenburg | Initial document                               |
+| 2020-11-26  | Wilfred Spiegelenburg | Allocation release definition, order change    |
+| 2020-11-27  | Wilfred Spiegelenburg | Sequence diagram                               |
+| 2020-12-12  | Wilfred Spiegelenburg | Application submit and cleanup                 |
+| 2020-12-15  | Wilfred Spiegelenburg | AllocationRelease message and cleanup sequence |
+| 2020-12-24  | Wilfred Spiegelenburg | Allocation changes for recovery                |
+| 2021-02-15  | Wilfred Spiegelenburg | Placeholder allocation and ask cleanup         |
+
+YUNIKORN-2 describes a new format for scheduling applications by taking into account the overall demand the application will have.
```
yangwwei, line 36: YUNIKORN-2 should be a link

yangwwei, line 36: overall demand -> minimal demand?

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+It guarantees the specified resources for the application by reserving the resources.
+
+There are two parts to this implementation:
+*   Kubernetes Shim
+*   Core and scheduling
+
+This document describes the implementation on the core side.
```
yangwwei, line 43: having a design doc for the shim and another one for the core is quite confusing
I think we should put them all in this doc
After I read all the doc, since it explains a lot of the expected behaviors of the shim, I don't think we should exclude K8shim implementation out of this doc, suggest removing this part.

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+
+| Change Date | Editors               | Description                                    |
+| ----------- | --------------------- | ---------------------------------------------- |
+| 2020-11-25  | Wilfred Spiegelenburg | Initial document                               |
+| 2020-11-26  | Wilfred Spiegelenburg | Allocation release definition, order change    |
+| 2020-11-27  | Wilfred Spiegelenburg | Sequence diagram                               |
+| 2020-12-12  | Wilfred Spiegelenburg | Application submit and cleanup                 |
+| 2020-12-15  | Wilfred Spiegelenburg | AllocationRelease message and cleanup sequence |
+| 2020-12-24  | Wilfred Spiegelenburg | Allocation changes for recovery                |
+| 2021-02-15  | Wilfred Spiegelenburg | Placeholder allocation and ask cleanup         |
```
yangwwei, line 34: To keep the format consistency, can we remove this part?
we can link to the original google doc if people want to know more about the changes

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+2. Scheduler storage object changes
+3. Scheduler logic changes
+
+## Non Goals
+Excluding the following major points:
+1. Kubernetes shim side implementation
+2. Generalised preemption on the core side
+
+## Generic flow
+As described in the YUNIKORN-2 design documentation we have the following flow.
```
yangwwei, line 58: what does "YUNIKORN-2 design documentation" refer to? It's better to be a link

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+Detailed information on the build up of the taskGroup(s), or the number of members are not relevant.
+The total resource requested by all taskGroup members is calculated using:
+
+![ask caclulation](./../assets/gang_total_ask.png)
+
+This total placeholderAsk is added as an optional field to the AddApplicationRequest message.
+The calculation can be made by the shim based on the CRD or annotation provided in the pod description.
+
+If the placeholderAsk is larger than the queue quota set on the queue the application must be rejected.
+This rejection is based on the fact that we cannot in any way honor the request For all other cases the application is accepted and will be scheduled as per normal.
```
yangwwei, line 108: minor: missing a period mark after " we cannot in any way honor the request"

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+
+The basic assumption is that all pods will generate a placeholder pod request to the core.
+This includes the pod that triggered the application creation if we do not use the application CRD.
+This assumption is needed to make sure that the scheduler core can behave in the same way for both ways of submitting the application.
+The placeholder pods must be communicated to the core before the real pod.
+
+Queue sorting for the queue that the application runs in must be set to _FIFO_ or _StateAware_.
+Other queue policies cannot guarantee that there is only one _New_ application processed at a time.
+In the case of the _FAIR_ policy we could be allocating multiple _New_ applications at the same time making quota management impossible to enforce.
+The other side effect of using _FAIR_ as a policy could be that we get multiple applications with only a partial allocated guarantee.
```
yangwwei, line 163: this seems to be overlapping with the previous section:
### Handling queue with a FAIR sort policy

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+6. Communicate the allocation back to the shim (if applicable, based on step 5)
+
+
+## Application completion
+Application completion has been a long standing issue.
+Currently, applications do not transition to a _completed_ state when done.
+The current states for the application are [documented here](./scheduler_object_states.md).
+However, at this point in time an application will not reach the _completed_ state and will be stuck in _waiting_.
+
+This provides a number of issues specifically around memory usage and cleanup of queues in long running deployments.
```
yangwwei, line 211: I think this needs to be updated. Now we have introduced `Completed`, `Completing`, `Failed`, and `Failing` state. So the app completion is no longer an issue. 
The second thing is: what does this sentence mean here?

```
However, at this point in time an application will not reach the _completed_ state and will be stuck in _waiting_.
```

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+
+### Definition
+Since we cannot rely on the application, running as pods on Kubernetes, to show that it has finished we need to define when we consider an application _completed_.
+At this point we are defining that an application is _completed_ when it has been in the _waiting_ state for a defined time period.
+An application enters the waiting state at the time that there are no active allocations (allocated resources > 0) and pending allocation asks (pending resources > 0).
+
+The transition to a _waiting_ state is already implemented.
+The time out of the _waiting_ state is new functionality.
```
yangwwei, line 219: Waiting state has been renamed to Completing
can we link this to the app state doc instead of explaining all these things again?

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+Placeholder asks are considered pending resource asks.
+These cases will be handled in the [Cleanup](#Cleanup) below.
+
+### Cleanup
+When we look at gang scheduling there is a further issue around unused placeholders, placeholder asks and their cleanup.
+Placeholders could be converted into real allocations at any time there are pending allocation asks or active allocations.
+
+Placeholder asks will all be converted into placeholder allocations before the real allocations are processed.
+
+Entry into the _waiting_ state is already handled.
```
yangwwei, line 231: Waiting now is Completing

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+...
+  // The name of the TaskGroup this ask belongs to
+  string taskGroupName = 10;
+  // Is this a placeholder ask (true) or a real ask (false), defaults to false
+  // ignored if the taskGroupName is not set
+  bool placeholder = 11;
+...
+}
+```
+
```
yangwwei, line 384: minor: extra empty line

docs/design/gang_scheduling.md
```diff
@@ -0,0 +1,631 @@
...
+Define the following points:
+
+1. Define changes required for the shim to core communication (scheduler interface)
+2. Scheduler storage object changes
+3. Scheduler logic changes
+
+## Non Goals
+Excluding the following major points:
+1. Kubernetes shim side implementation
+2. Generalised preemption on the core side
```
yangwwei, line 55: The Gaols and Non Goals here seems to be what this doc wants to elaborate, instead of what we want to achieve with the gang scheduling feature.. this seems to be a bit odd.

yangwwei (changes requested):

yangwwei:
One thing I do not find in this doc is:  how we handle the container SWAP? 
I think that is also very important, it will be great to add some doc to explain that as well.

docs/design/gang_scheduling.md
wilfred-s: removed

docs/design/gang_scheduling.md
wilfred-s: removed

docs/design/gang_scheduling.md
wilfred-s: That was done after design was written. We should not change the design doc to fit the implementation afterwards.

docs/design/gang_scheduling.md
wilfred-s: no, this is not the user guide this is a design doc as it was written to design and implement the feature. We should not change the design doc to fit the implementation afterwards.

docs/design/gang_scheduling.md
wilfred-s: see above

docs/design/gang_scheduling.md
wilfred-s: text removed

docs/design/gang_scheduling.md
wilfred-s: Did you write a design for the shim after the major change to the design in November? The last doc version that I saw for the shim design was from before that time. In that version of the design the shim handled the placeholder allocation and the core was not aware of placeholders at all.
If we have a design that we implemented, or an updated design, that at least aligns with the implementation we can add it to this doc and update it.

docs/design/gang_scheduling.md
wilfred-s: removed, cleaned up other empty lines too

docs/design/gang_scheduling.md
wilfred-s: text updated

wilfred-s:
> One thing I do not find in this doc is: how we handle the container SWAP?
> I think that is also very important, it will be great to add some doc to explain that as well.

The exact implementation of the swap is not given. The sequence of how the swap works and what triggers what is shown in the generic flow diagram at steps (7) and (8). Included in the text also.
We also cover the swap in the scheduler logic changes, the last step list before application completion.

docs/design/gang_scheduling.md
yangwwei: Agree, we can add that into this doc later.

docs/design/gang_scheduling.md
yangwwei: @wilfred-s  can you update this? 
if you like, you can remove it

docs/design/gang_scheduling.md
yangwwei: Not the exact point, but still part of the design. Having the design doc saying something but the implementation does something else seems to be a problem to me.

docs/design/gang_scheduling.md
wilfred-s: copy paste error, fixed

docs/design/gang_scheduling.md
wilfred-s: fixed

docs/design/gang_scheduling.md
wilfred-s: The failing & completing state were introduced to fix issues in the cleanup. That was done on top of what we had designed. We also do not go back and now fix the gang scheduling design after fixing YUNIKORN-639 which also changes the flow of the states.
A design you write, implement and then on top of the design you fix issues. You do not go back then and re-write the design doc. You update the user docs. There is a clear distinction between the two.

docs/design/gang_scheduling.md
wilfred-s: I think we need to leave this until we have the missing bits from the shim added and at that point we remove this.

docs/design/gang_scheduling.md
yangwwei: IMO, we need to keep this updated if the design changes.
Otherwise, what's the point to have an out-of-dated doc here, people will get confused when they look into this doc and trying to understand what has been implemented.

yangwwei (approved):
+1
lets update this doc with a follow up JIRA https://issues.apache.org/jira/browse/YUNIKORN-648 to make sure it is up-to-date

PR #52, opened 2021-05-11
[YUNIKORN-651] Add document to explain how to define user identity

avsamit6600:
This PR is specifically to modify all user references (except ACLs). The ACLs will be modified as part of this JIRA(https://issues.apache.org/jira/browse/YUNIKORN-652)

### DIFF
--- a/docs/user_guide/usergroup_resolution.md
+++ b/docs/user_guide/usergroup_resolution.md
@@ -1,0 +1,65 @@
+---
+id: usergroup_resolution
+title: User & Group Resolution
+---
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+## User resolution
+
+User information is an important aspect of the scheduling cycle. It is one of the key identifier that can be used to determine the queue to which a job should be submitted. The Yunikorn Scheduler relies on the K8s Shim to provide user information. In the world of Kubernetes, there is no object defined that identfies the actual user. This is by design and more information can be found [here](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#users-in-kubernetes)
+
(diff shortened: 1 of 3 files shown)


docs/user_guide/usergroup_resolution.md
```diff
@@ -0,0 +1,65 @@
...
+  labels:
+    yunikorn.apache.org/username: "john"
+```
+NOTE: 
+In order to make this field uniquiely identifiable to the authorized user, the suggestion is to add this label as an immutable field by the user identification tool used by the cluster administrators. The cluster administrators or users are free to use any method or tool to add this field and value. This includes adding it manually at the time of submission. 
+
+Assumption:
+  Yunikorn assumes that all pods belonging to an application are owned by the same user. We recommend that the user label is added to every pod of an app. This is to ensure that there is no discrepency. 
+
+The `yunikorn.apache.org/username` key can customized by overriding the default value using the `USER_LABEL_KEY`env variable to the [K8s](https://github.com/apache/incubator-yunikorn-release/blob/744ce13ec88c49596f965ad0908c25f555cd356c/helm-charts/yunikorn/templates/deployment.yaml#L48). This is particularly useful in scenarios where the user label is already being added or if the label has to be modified for some secuirty reasons.
```
yangwwei, line 47: minor typo:
can customized -> can be customized

yangwwei, line 47: Can we point this to the master version: https://github.com/apache/incubator-yunikorn-release/blob/master/helm-charts/yunikorn/templates/deployment.yaml? Ideally we should allow this to be configurable via our helm charts. do you want to create another JIRA for this?

yangwwei (changes requested):

docs/user_guide/usergroup_resolution.md
```diff
@@ -0,0 +1,65 @@
...
+| yunikorn.apache.org/username 	                 | User name. It can have duplicate entries but only the first value will be used. The default user is `nobody` |
+
+Example:
+```yaml
+metadata:
+  labels:
+    yunikorn.apache.org/username: "john"
+```
+NOTE: 
+In order to make this field uniquiely identifiable to the authorized user, the suggestion is to add this label as an immutable field by the user identification tool used by the cluster administrators. The cluster administrators or users are free to use any method or tool to add this field and value. This includes adding it manually at the time of submission.
```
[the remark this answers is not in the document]
avsamit6600: Sure. I will add this as a TIP.

docs/user_guide/usergroup_resolution.md
```diff
@@ -0,0 +1,65 @@
...
+```yaml
+metadata:
+  labels:
+    yunikorn.apache.org/username: "john"
+```
+NOTE: 
+In order to make this field uniquiely identifiable to the authorized user, the suggestion is to add this label as an immutable field by the user identification tool used by the cluster administrators. The cluster administrators or users are free to use any method or tool to add this field and value. This includes adding it manually at the time of submission. 
+
+Assumption:
+  Yunikorn assumes that all pods belonging to an application are owned by the same user. We recommend that the user label is added to every pod of an app. This is to ensure that there is no discrepency.
```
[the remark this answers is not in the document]
avsamit6600: Cool. Modifying this as a note with "Assumption" as the title

docs/user_guide/usergroup_resolution.md
avsamit6600: changed. thanks

avsamit6600: Yes, created another JIRA for this:
https://issues.apache.org/jira/browse/YUNIKORN-667

yangwwei (approved):
LGTM

PR #54, opened 2021-05-21
YUNIKORN-515: Documentation changes for all Redesigned API's

manirajv06:
Doc changes for all redesigned API's and describing json error responses for other API's as well.

### DIFF
--- a/docs/api/scheduler.md
+++ b/docs/api/scheduler.md
@@ -22,7 +22,159 @@ specific language governing permissions and limitations
 under the License.
 -->
 
-## Queues
+## Partitions
+
+Displays general information about the partition like name, state, capacity, used capacity and node sorting policy. 
+
+**URL** : `/ws/v1/partitions`
+
+**Method** : `GET`
+
+**Auth required** : NO
+
+### Success response
+
+**Code** : `200 OK`
+
+**Content examples**
+
+```json
+[
+    {
+        "name": "[mycluster]default",
+        "state": "Active",
+        "lastStateTransitionTime": "2021-05-20 12:25:49.018953 +0530 IST m=+0.005949717",
+        "capacity": {
+            "capacity": "[memory:1000 vcore:1000]",
+            "usedcapacity": "[memory:800 vcore:500]"
+        },
+        "nodeSortingPolicy": "fair",
+        "applications": {
+            "New": 5,
+            "Pending": 5,
+            "total": 10
+        }
+    },
+    {
+        "name": "[mycluster]gpu",
+        "state": "Active",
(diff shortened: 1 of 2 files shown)


docs/api/scheduler.md
```diff
@@ -22,7 +22,159 @@ specific language governing permissions and limitations
...
+}
+```
+
+```json
+{
+    "status_code": 500,
+    "message": "system error message. for example, json: invalid UTF-8 in string: ..",
+    "description": "system error message. for example, json: invalid UTF-8 in string: .."
+}
+```
```
yangwwei, line 175: I don't think we need to give so many examples
Just one example with some explanation should be enough, such as:

```
{    
    "status_code": 500,
    "message": "error message", 
    "description": "explanation about this error"
}
```

docs/api/scheduler.md
```diff
@@ -22,7 +22,159 @@ specific language governing permissions and limitations
...
+               "parent": "root"
+            }
+        ]
+    } 
+]
+```
+
+### Error response
+
+**Code** : `500 Internal Server Error`
```
yangwwei, line 141: Do we only return 500 error code?
I think in the code sometimes we also return BadRequest (400)

docs/api/scheduler.md
```diff
@@ -77,7 +229,137 @@ For the default queue hierarchy (only `root.default` leaf queue exists) a simila
...
+
+```json
+{
+    "status_code": 500,
+    "message": "system error message. for example, json: invalid UTF-8 in string: ..",
+    "description": "system error message. for example, json: invalid UTF-8 in string: .."
+}
+```
+
+## Applications (Deprecated)
```
yangwwei, line 362: Same as above, I do not think we need so many example content
Just one should be enough, try to keep the doc neat

docs/api/scheduler.md
```diff
@@ -159,7 +441,126 @@ In the example below there are three allocations belonging to two applications.
...
+}
+```
+
+```json
+{
+    "status_code": 500,
+    "message": "system error message. for example, json: invalid UTF-8 in string: ..",
+    "description": "system error message. for example, json: invalid UTF-8 in string: .."
+}
+```
```
yangwwei, line 561: same as above

docs/api/scheduler.md
```diff
@@ -469,6 +884,112 @@ Reponse
...
```
yangwwei, line 991: same as above

yangwwei (changes requested):

yangwwei (approved):
LGTM

PR #59, opened 2021-06-24
[YUNIKORN-726] Add instructions to approve cert requests

holdenk:
These requests may not be auto-approved in newer version of Kubernetes

### DIFF
--- a/docs/get_started/get_started.md
+++ b/docs/get_started/get_started.md
@@ -43,6 +43,14 @@ By default, the helm chart will install the scheduler, web-server and the admiss
 When `admission-controller` is installed, it simply routes all traffic to YuniKorn. That means the resource scheduling
 is delegated to YuniKorn. You can disable it by setting `embedAdmissionController` flag to false during the helm install.  
 
+
+If the admission controller enters a crashloopbackoff you may need to manually approve the certificate request.
+
+```shell script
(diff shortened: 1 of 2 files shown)


wilfred-s:
This is the line that approves the request in the [script](https://github.com/apache/incubator-yunikorn-k8shim/blob/0bd93660ec8b9b99dd28ae93ea755a52d0187741/deployments/admission-controllers/scheduler/generate-signed-ca.sh#L91)
Is this related to YUNIKORN-625 and indirectly YUNIKORN-674? If so we should work on that issue. I have mentioned in discussion before that we should use go code and not a shell script to deploy the admin controller. That would allow us to recompile the admin deployment code for specific K8s versions.

wilfred-s:
Based on the changes we have made there is no longer a script and we no longer need to approve any certificates.
@holdenk I think this has been made obsolete with the new deployment code of the admission controller. 
I think we can close this one, please confirm if that is correct.

craigcondit:
> Based on the changes we have made there is no longer a script and we no longer need to approve any certificates.

Correct.  

> @holdenk I think this has been made obsolete with the new deployment code of the admission controller.

Correct. 

> I think we can close this one, please confirm if that is correct.

Yes we can close this one.

craigcondit:
Closing, as this is no longer needed.

PR #60, opened 2021-06-24
[YUNIKORN-727] Fix Spark RBAC in workspace examples

holdenk:
We need to specify the namespace for the subjects since the -n doesn't propegate the NS to subject. Might as well put the SA inside of the NS while were at it.


yangwwei (approved):
+1, thanks for fixing this @holdenk

PR #70, opened 2021-08-12
[YUNIKORN-779] fix commands on download page

wilfred-s:
The command mentioned on the download page have non ascii chars embedded.
It also provides the wrong command for the checksum validation.


yangwwei (approved):
+1

PR #73, opened 2021-08-20
[YUNIKORN-808] document child template

chia7712:
issue: https://issues.apache.org/jira/browse/YUNIKORN-808


docs/user_guide/queue_config.md
```diff
@@ -302,3 +302,45 @@ resources:
...
+             childtemplate:
+               resources:
+                 max:
+                   vcore: 21000
+                   memory: 610000
+             queues:
+               - name: a00
+```
+In this case, the dynamic leaf (`root.b0`) uses the template of `root` to initialize `application.sort.policy`, max resource, and guaranteed resource.
+By contrast, the dynamic leaf (`root.a0.a10`) uses the template of `root.a0` (the closest parent) to initialize configurations.
```
chia7712, line 342: this behavior will get fixed by https://github.com/apache/incubator-yunikorn-core/pull/308

docs/user_guide/queue_config.md
```diff
@@ -302,3 +302,45 @@ resources:
...
 ```
 Resources that are not specified in the list are not limited, for max resources, or guaranteed in the case of guaranteed resources. 
+
+### Child Template
+
+The parent queue can provide a template to define the behaviour of dynamic leaf queues below it. The supported configuration in template are shown below.
+1. application sort policy
+2. max resources
+3. guaranteed resources
+4. max application
```
wilfred-s, line 312: This one is missing from the example below.

docs/user_guide/queue_config.md
```diff
@@ -302,3 +302,45 @@ resources:
...
+         submitacl: '*'
+         childtemplate:
+           properties:
+             application.sort.policy: stateaware
+           resources:
+             max:
+               vcore: 20000
+               memory: 600000
+         queues:
+           - name: a0
```
wilfred-s, line 332: can we use a more logical name like `parent`?

docs/user_guide/queue_config.md
```diff
@@ -302,3 +302,45 @@ resources:
...
+             childtemplate:
+               resources:
+                 max:
+                   vcore: 21000
+                   memory: 610000
+             queues:
+               - name: a00
```
wilfred-s, line 339: This queue is unused, better to add a `parent: true` tag to the `root.a0` queue (rename that as per the above comment)

docs/user_guide/queue_config.md
```diff
@@ -302,3 +302,45 @@ resources:
...
+```
+In this case, the dynamic leaf (`root.b0`) uses the template of `root` to initialize `application.sort.policy`, max resource, and guaranteed resource.
```
wilfred-s, line 341: In this case, the dynamic leaf queue `root.sales`, will use the template set on the parent queue, in this case `root` to....

docs/user_guide/queue_config.md
wilfred-s: This is the expected behaviour so that is good to document.

Explain how inheritance works at this point: a parent queue inherits the child template from its parent if that parent has one defined. A child template can be defined at any level in the queue hierarchy on a queue that is of the type `parent`

Use more logical queue names to make the example clearer

wilfred-s (changes requested):

chia7712:
@wilfred-s thanks for all suggestions. I have updated the PR. PTAL.

wilfred-s (approved):
changes look good

PR #79, opened 2021-09-06
[YUNIKORN-835] Reminder readers that web-latest image in docker hub i…

chia7712:
issue: https://issues.apache.org/jira/browse/YUNIKORN-835

### DIFF
--- a/docs/developer_guide/deployment.md
+++ b/docs/developer_guide/deployment.md
@@ -37,6 +37,7 @@ This command will build an image. The image will be tagged with a default versio
 
 **Note** the default build uses a hardcoded user and tag. You *must* update the `IMAGE_TAG` variable in the `Makefile` to push to an appropriate repository. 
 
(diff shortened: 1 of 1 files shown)


docs/developer_guide/deployment.md
```diff
@@ -37,6 +37,7 @@ This command will build an image. The image will be tagged with a default versio
 
 **Note** the default build uses a hardcoded user and tag. You *must* update the `IMAGE_TAG` variable in the `Makefile` to push to an appropriate repository. 
 
+**Note** the yunikorn docker images in docker hub are not updated anymore due to ASF policy. Hence, you should build both scheduler image and web image locally before deploying them.
```
kingamarton, line 40: I think we should mention that the `latest` images are not updated, just to avoid any confusions.

kingamarton (changes requested):

docs/developer_guide/deployment.md
chia7712: make sense. will copy that

kingamarton (approved):

PR #81, opened 2021-09-17
[YUNIKORN-852] Document resource weighting in node sorting policies.

craigcondit:


craigcondit:
@yangwwei, can you review this?

docs/user_guide/sorting_policies.md
```diff
@@ -146,6 +146,33 @@ Resources of the same type are compared for the nodes.
 This results in a node with the highest utilisation to be considered first for assigning new allocation.
 Resulting in a high(er) utilisation of a small(er) number of nodes, better suited for cloud deployments.   
 
+## Resource weighting
+Node sorting policies may use the utilization of a node to determine ordering. Because nodes can have several unique
+resource types, a node's utilization is determined by a weighted average of its individual resource types. Resource
+weighting can be customized by using the `resourceweights` section of `nodesortpolicy`. If `resourceweights` is not
+present or empty, the default configuration sets the weight of both `vcore` and `memory` equally to `1.0`. All other
+resource types are ignored. Only resource types explicitly mentioned will have a weight.
```
yangwwei, line 154: I think we need to add a note here, users might be confused about "vcore". We can say something like:
> YuniKorn tracks CPU resources as "vcores" internally, this is mapped to the K8s resource type "cpu". For any other resource types in K8s, the naming is consistent in YuniKorn.

docs/user_guide/sorting_policies.md
```diff
@@ -146,6 +146,33 @@ Resources of the same type are compared for the nodes.
...
+    nodesortpolicy:
+      resourceweights:
+        vcore: 4.0
+        memory: 1.0
+```
+
+With this configuration, In this example, if a node has `90%` of its CPU and `50%` of its memory allocated, the node
+will be considered to be `82%` utilized.
+
+Note that weights are relative to each other, so specifying weights of `{ 4.0, 1.0 }` is equivalent to
```
yangwwei, line 173: can we mention negative weight is not allowed?

docs/user_guide/sorting_policies.md
```diff
@@ -146,6 +146,33 @@ Resources of the same type are compared for the nodes.
...
+
+The following configuration entry sets the weight of `vcore` to `4.0` and `memory` to `1.0`. This will weight CPU
+usage four times higher than memory usage:
+```yaml
+partitions:
+  - name: default
+    nodesortpolicy:
+      resourceweights:
+        vcore: 4.0
+        memory: 1.0
```
yangwwei, line 167: can we include `type` in this example as well?
I think when user configs `nodesortingpolicy`, type should be always provided.

yangwwei (changes requested):

docs/user_guide/sorting_policies.md
craigcondit: Updated.

docs/user_guide/sorting_policies.md
craigcondit: Updated.

docs/user_guide/sorting_policies.md
craigcondit: Updated.

yangwwei (approved):
+1

PR #88, opened 2021-11-06
[YUNIKORN-864] update design documentation with new interface

manirajv06:
update scheduler core design documentation with new interface

### DIFF
--- a/docs/design/interface_message_simplification.md
+++ b/docs/design/interface_message_simplification.md
@@ -1,0 +1,309 @@
+---
+id: interface_message_simplication
+title: Simplyfing Interface Messages
+---
+
+<!--
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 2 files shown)


wilfred-s:
Based on the comment in the top of the document I think we should not update the `scheduler_core_design.md` at all: [caution](https://github.com/apache/incubator-yunikorn-site/blob/master/docs/design/scheduler_core_design.md?plain=1#L25)
The messages processing has been simplified with the removal of the scheduler cache and most, if not all, of the document is obsolete. We should really start work on YUNIKORN-495 to get this fixed.

manirajv06:
Ok. Let me give a try on YUNIKORN-495

manirajv06:
@wilfred-s as discussed offline, created a new doc to describe interface message simplification, code flow analysis based on message changes, breaking shim dependency etc. Please take a look. Yes, we will need to revert previous commit done on Scheduler Cache design doc and we may need to add a caution header to Scheduler cache removal design doc to refer to this new doc.

wilfred-s:
I have reverted the file that we do not maintain anymore.

wilfred-s (approved):
LGTM

PR #91, opened 2021-11-16
[YUNIKORN-944] Fix build dependencies

wilfred-s:
The search theme requires a later version of node.js to build.
Fix the local build script and the README to add the search dependency.
Update node to the current release 16.13.
Upgrade docusaurus to the latest v2 beta release (moving away from the
current alpha release)


yuchaoran2011 (approved):
LGTM

PR #101, opened 2021-12-10
[YUNIKORN-845] Publish benchmark result on web-site

0yukali0:
Update performance docs in v0.12.0
Screenshot from 2021-12-10 16-05-47


0yukali0:
Hi @yangwwei 
I copied master diagram and performance docs to v0.12.0 docs

yangwwei (approved):
Looks good. +1
btw, I have reverted the change to package.json, that updates a version which is not needed for this patch.

PR #104, opened 2022-01-06
[YUNIKORN-988] replace navbar image

wilfred-s:
The navbar image used is an ico file which does not correctly get
rendered on mobile devices. Replace the image with a png version.

Upgrade docusaurus to the latest beta version.


yangwwei (approved):
thanks for fixing this @wilfred-s

yuchaoran2011 (approved):
It's probably not convenient to see how this change will look like on mobile without merging it

PR #113, opened 2022-01-25
[YUNIKORN-1044] Publish design doc for scheduler plugin

craigcondit:
https://issues.apache.org/jira/browse/YUNIKORN-1044

### DIFF
--- a/docs/design/scheduler_plugin.md
+++ b/docs/design/scheduler_plugin.md
@@ -92,7 +92,7 @@ non-batch workloads.
 If an `applicationId` is present, then we determine if there is a pending pod allocation (meaning the
 YuniKorn core has already decided to allocate the pod). If so, we return `Success`, otherwise `Unschedulable`.
 Additionally, if an in-progress allocation is detected (indicating that we have previously attempted to
-schedule this pod), we trigger a `TaskFailed` event for the YuniKorn core so that the pod will be sent back
+schedule this pod), we trigger a `RejectTask` event for the YuniKorn core so that the pod will be sent back
 for scheduling later.
 
 ### Filter
--- a/docs/design/scheduler_plugin.md
+++ b/docs/design/scheduler_plugin.md
@@ -1,0 +1,112 @@
+---
+id: scheduler_plugin
+title: K8s Scheduler Plugin
+---
+
+<!--
+* Licensed to the Apache Software Foundation (ASF) under one
+* or more contributor license agreements.  See the NOTICE file
+* distributed with this work for additional information
+* regarding copyright ownership.  The ASF licenses this file
+* to you under the Apache License, Version 2.0 (the
+* "License"); you may not use this file except in compliance
+* with the License.  You may obtain a copy of the License at
+*
+*      http://www.apache.org/licenses/LICENSE-2.0
+*
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+-->
+
+## Background
+
+YuniKorn (on Kubernetes) is traditionally implemented as a ground-up implementation of a Kubernetes scheduler.
+This has allowed us to innovate rapidly, but is not without its problems; we currently have numerous places
+where we call into non-public K8S source code APIs with varying levels of (code) stability, requiring
+sometimes very disruptive code changes when we switch to new Kubernetes releases.
+
+Ideally, we should be able to take advantage of enhancements to new Kubernetes releases automatically.
+Using the plugin model enables us to enhance the Kubernetes scheduling logic with YuniKorn features.
+This also helps keep YuniKorn compatible with new Kubernetes releases with minimal effort.
+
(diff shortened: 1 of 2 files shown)


docs/design/scheduler_plugin.md
```diff
@@ -0,0 +1,112 @@
...
+functionality and use default scheduling logic. However, we have no way to do that today as the default
+scheduling functionality is not present in the YuniKorn scheduler binary.
+
+Since Kubernetes 1.19, the Kubernetes project has created a stable API for the
+[Scheduling Framework](https://kubernetes.io/docs/concepts/scheduling-eviction/scheduling-framework/),
+which allows plugins to be created which implement various extension points. Plugins implement one or more
+of these extension points, and are then compiled into a scheduler binary which contains the default
+scheduler and plugin code, configured to call into the plugins during normal scheduling flow.
+
+## Design
```
yangwwei, line 45: Some things might be good to be included:
1. We need a high-level design to explain how this works together. A workflow will be helpful. For example, when a pod gets submitted, what happens in the scheduler-plugin mode until it gets allocated?
2. Can you help to list the features that cannot be supported in the plugin mode?
3. I think we need to mentioned the plugin mode will have worse performance than the standalone mode

docs/design/scheduler_plugin.md
```diff
@@ -0,0 +1,112 @@
...
+## Plugin Implementation
+
+To expose the entirety of YuniKorn functionality, we implement three of the Scheduling Framework Plugins:
+
+### PreFilter
+
+PreFilter plugins are passed a reference to a Pod and return either `Success` or `Unschedulable`, depending
+on whether that pod should be considered for scheduling.
+
+For the YuniKorn implementation, we first check the Pod to see if we have an associated `applicationID`
```
yangwwei, line 88: applicationID -> applicationId

docs/design/scheduler_plugin.md
```diff
@@ -0,0 +1,112 @@
...
+defined. If not, we immediately return `Success`, which allows us to delegate to the default scheduler for
+non-batch workloads.
+
+If an `applicationId` is present, then we determine if there is a pending pod allocation (meaning the
+YuniKorn core has already decided to allocate the pod). If so, we return `Success`, otherwise `Unschedulable`.
+Additionally, if an in-progress allocation is detected (indicating that we have previously attempted to
+schedule this pod), we trigger a `TaskFailed` event for the YuniKorn core so that the pod will be sent back
+for scheduling later.
```
yangwwei, line 96: I found this is quite confusing. If we have previously attempted to schedule a pod, and in PreFilter we see it again, we directly failed the allocation. Why do we consider this a failure? What if the actual binding took some time and is still ongoing? 

In another word, under what circumstance we can see a pod again in the PreFilter() plugin which was already scheduled by the default scheduler?

yangwwei (changes requested):
And it looks like we do not have a user doc about how to build, deploy and use the scheduler plugin, can we add that in a separate JIRA?

docs/design/scheduler_plugin.md
craigcondit: > 1. We need a high-level design to explain how this works together. A workflow will be helpful. For example, when a pod gets submitted, what happens in the scheduler-plugin mode until it gets allocated?
As the document indicates, YuniKorn runs as normal in the background with very little change (it just defers the actual binding until the scheduler framework calls `PreFilter()` / `Filter()`. I can take a stab at a diagram (though my illustration skills are not that great), but I'm not really sure what it would convey.

> 2. Can you help to list the features that cannot be supported in the plugin mode?
N/A, as there are no unsupported features with the plugin mode.

> 3. I think we need to mentioned the plugin mode will have worse performance than the standalone mode
This has yet to be determined, and I'm not sure it's actually the case. I'm not going to speculate about it here.

docs/design/scheduler_plugin.md
craigcondit: If the default scheduler is unable to bind the pod, it will attempt to reschedule it, so we will see the same pod again in PreFilter(). In this case, we need to ensure that we release all the allocations in the core and send it back to the start, otherwise we risk being unable to schedule the pod at all. Note that this will only happen if the scheduler cannot bind a pod after it has already been cleared by Filter().

docs/design/scheduler_plugin.md
craigcondit: done

craigcondit:
> And it looks like we do not have a user doc about how to build, deploy and use the scheduler plugin, can we add that in a separate JIRA?

Yes, we should add this in a follow-up.

docs/design/scheduler_plugin.md
yangwwei: Is this tested? If we mark the task as failed, I don't think the task will be rescheduled in YK.

yangwwei (approved):

docs/design/scheduler_plugin.md
craigcondit: updated to reflect that we send 'RejectTask' instead

yangwwei: After some long discussion. The current doc is aligning with the current implementation. But for this part, especially for retry logic, we need to do more tests and ensure the behavior is correct. If in the future we find anything incorrect that needs to be fixed, we will revisit the doc. Thx

PR #115, opened 2022-01-27
[YUNIKORN-1056] Document usage of scheduler plugin

craigcondit:
https://issues.apache.org/jira/browse/YUNIKORN-1056

### DIFF
--- a/docs/developer_guide/deployment.md
+++ b/docs/developer_guide/deployment.md
@@ -98,6 +98,11 @@ The deployment will run 2 containers from your pre-built docker images in 1 pod,
 * yunikorn-scheduler-core (yunikorn scheduler core and shim for K8s)
 * yunikorn-scheduler-web (web UI)
 
+Alternatively, the scheduler can be deployed as a K8S scheduler plugin:
+```
+kubectl create -f deployments/scheduler/plugin.yaml
+```
+
 The pod is deployed as a customized scheduler, it will take the responsibility to schedule pods which explicitly specifies `schedulerName: yunikorn` in pod's spec. In addition to the `schedulerName`, you will also have to add a label `applicationId` to the pod.
 ```yaml
   metadata:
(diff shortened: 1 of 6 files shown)


docs/developer_guide/deployment.md
```diff
@@ -98,6 +98,11 @@ The deployment will run 2 containers from your pre-built docker images in 1 pod,
 * yunikorn-scheduler-core (yunikorn scheduler core and shim for K8s)
 * yunikorn-scheduler-web (web UI)
 
+Alternatively, the scheduler can be deployed as a K8S scheduler plugin:
+```
+kubectl create -f deployments/scheduler/plugin.yaml
+```
+
```
yangwwei, line 105: ## Deploy the Scheduler

There are 2 deployment modes for yunikorn, standalone and scheduler-plugin. In standalone mode, yunikorn runs as a customized K8s scheduler and is supposed to replace the default scheduler in the cluster; in the scheduler-plugin mode, yunikorn is built based on K8s scheduling framework as a scheduler-plugin, preserves all default scheduler functionalities plus the yunikorn features. 

### Standalone
// doc to deploy as the standalone scheduler

### scheduler-plugin
// doc to deploy as a plugin

yangwwei (changes requested):
hi @craigcondit thanks for adding this. This is for the developer. Can we add some doc in the user guide as well? Essentially we should have a page to explain what's are the 2 different modes, standalone vs plugin. And what does that mean to the user? I do not think we have anywhere explained this in detail today, which is a major thing missing.

docs/developer_guide/deployment.md
craigcondit: @yangwwei , that's in here as well under the getting started page.

craigcondit: Added a new comparision page that contrasts the deployment modes.

yangwwei: Thanks

yangwwei (approved):
LGTM

PR #120, opened 2022-01-28
[YUNIKORN-962] Add deprecation message to REST calls.

wilfred-s:
The REST end points `/ws/v1/queues`, `/ws/v1/apps` and `/ws/v1/nodes`
have been deprecated.


craigcondit (approved):
+1.

PR #127, opened 2022-02-05
[YUNIKORN-1068] Doc changes for state dump file path

anuraagnalluri:


anuraagnalluri:
FYI @manirajv06

docs/user_guide/queue_config.md
```diff
@@ -191,7 +192,18 @@ partitions:
 The placement rules are defined and documented in the [placement rule](placement_rules.md) document.
 
 Each partition can have only one set of placement rules defined. 
-If no rules are defined the placement manager is not started and each application *must* have a queue set on submit.  
+If no rules are defined the placement manager is not started and each application *must* have a queue set on submit.
+
+### Statedump filepath
+
+The statedump filepath defines the output file for YuniKorn statedumps. It is optionally set on the partition level. If set,
```
yangwwei, line 199: NIT: defines the output files for YuniKorn state dumps

there can be more than one file, correct? Do we need to briefly talk about the rotation policy here as well?

docs/user_guide/queue_config.md
```diff
@@ -191,7 +192,18 @@ partitions:
...
+the value of this field can be either a relative (to working directory) or absolute path. YuniKorn scheduler will error if it does not have sufficient permissions to create the statedump file at the specified path.
```
yangwwei, line 200: > YuniKorn scheduler will error if it does not have sufficient

NIT: YuniKorn scheduler will be unable to start if it does not have sufficient

yangwwei (changes requested):

docs/user_guide/queue_config.md
anuraagnalluri: Thanks for pointing that out -- added details about the rotation policy below.

There is only one file that is being written to for statedumps as per the log rotation policy. Different partitions can technically have different values, but statedumps contain information across multiple partitions. As a result, only the first occurrence of `statedumpfilepath` takes precedence for defining the location of the output file, even if other partitions are configured with different values. If no occurrence is found, it will default to `yunikorn-state.txt` (previous behavior).

yangwwei: Thanks for the clarification, it helps!

yangwwei (approved):
LGTM.

yangwwei:
Merging this, thank you @anuraagnalluri for getting this done.
BTW, please do not use force push to update your PR, it will be easier for the reviewer to check the commit history in order to figure out the diff between them.

PR #129, opened 2022-02-12
[YUNIKORN-990] DocSearch Migration

0yukali0:
#Description
Update credentials which are new apiKey and appId of algolia docsearch, according to [Docsearch migration in docusaurus blog](https://docusaurus.io/blog/2021/11/21/algolia-docsearch-migration).

#Build and run(test)
image


yangwwei (approved):
thanks for getting this done.
I did not know that the migration is just about the appId and apiKey update 👍

PR #151, opened 2022-04-06
[YUNIKORN-596]document pod labels and annotations

HuangTing-Yao:
Add pod labels and annotation introduction.


craigcondit (approved):
+1 other than some grammatical errors. Will fix on commit.

PR #164, opened 2022-07-12
[YUNIKORN-1225] pin docusaurus version

wilfred-s:
Pin the docusaurus version to not change in the package.json
Update react to the latest supported version for Docusaurus
Move to .nvmrc for the node version for local and push builds


0yukali0 (approved):
LGTM.
I tried local-build.sh run and build on Mac and this commit worked.
There was not additional modified lines in package.json after finishing the command was.

PR #178, opened 2022-08-23
[YUNIKORN-1289] Publish Daemonset scheduling design doc

manirajv06:


docs/design/simple_preemptor.md
```diff
@@ -0,0 +1,114 @@
+             ---
```
wilfred-s, line 1: No spaces at the start of the line

sidebars.js
```diff
@@ -75,6 +75,7 @@ module.exports = {
                     'design/state_aware_scheduling',
                     'design/scheduler_object_states',
                     'design/scheduler_core_design',
+                    'design/simple_preemptor'
```
wilfred-s, line 78: This should not be in between the two old/no longer maintained docs. It should be higher up the list, around the generic resource line

wilfred-s (changes requested):
two minor point to fix

craigcondit (approved):
+1 LGTM

PR #180, opened 2022-08-25
[YUNIKORN-1293] Add custom redirects to the current version doc

yangwwei:
Docusaurus hides the version number in the URL for the latest versions, e.g right now our latest version is 1.0.0, but http://localhost:3000/docs/1.0.0 gives a 404. Ideally, we should make have an accessible URL for this version as well. This is in order to fix: https://github.com/apache/spark/pull/37622

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -36,6 +36,19 @@ module.exports = {
       },
     },
   },
+  plugins: [
+    [
+      '@docusaurus/plugin-client-redirects',
+      {
+        redirects: [
+          {
+            to: '/docs/1.0.0',
+            from: '/docs/',
+          },
+        ],
+      },
+    ],
+  ],
   presets: [
     [
       '@docusaurus/preset-classic',
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@docusaurus/core": "2.0.1",
+    "@docusaurus/plugin-client-redirects": "^2.0.1",
     "@docusaurus/preset-classic": "2.0.1",
     "@docusaurus/theme-search-algolia": "^2.0.1",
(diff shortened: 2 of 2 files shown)


yuchaoran2011 (approved):

dongjoon-hyun (commented):
Thank you, @yangwwei !

wilfred-s:
This is not a scalable or maintainable solution. It breaks the release procedure.

Not sure why YuniKorn needs to comply with this rule for a specific version of documentation when there is no version information in any of the volcano links published as part of the [Spark 3.3 documentation](https://spark.apache.org/docs/latest/running-on-kubernetes.html#using-volcano-as-customized-scheduler-for-spark-on-kubernetes):
* https://volcano.sh/en/docs/podgroup/
* https://volcano.sh/en/docs/podgroup/#example

yangwwei:
> This is not a scalable or maintainable solution. It breaks the release procedure.
> 
> Not sure why YuniKorn needs to comply with this rule for a specific version of documentation when there is no version information in any of the volcano links published as part of the [Spark 3.3 documentation](https://spark.apache.org/docs/latest/running-on-kubernetes.html#using-volcano-as-customized-scheduler-for-spark-on-kubernetes):
> 
> * https://volcano.sh/en/docs/podgroup/
> * https://volcano.sh/en/docs/podgroup/#example

This is a very good point. Let's move the discussion to the Spark community: https://github.com/apache/spark/pull/37622. This PR is not a sustainable solution, it's gonna break when we release a new version. We can revert this once we reach a consensus there.  BTW, I've spent hours and this is the best solution so far that can work.

wilfred-s:
We'll leave it in place for the next days, release 1.1 is due next week and should include a new versioned release for the docs to be created. Before we can commit that update this change will most likely need to be rolled back to not break our docs.

PR #185, opened 2022-09-09
[YUNIKORN-1312] update website for 1.1.0

pbacsko:


package.json
```diff
@@ -16,6 +16,7 @@
     "@docusaurus/theme-search-algolia": "^2.0.1",
     "@mdx-js/react": "^1.5.8",
     "clsx": "^1.1.1",
+    "node": "^18.8.0",
```
pbacsko, line 19: This was put here by yarn 1.22.19.

wilfred-s (changes requested):
missing 
* update for the download page 
* announcementBar version in the docusaurus.config.js needs an update

package.json
wilfred-s: We build using a docker image so I am not sure that this will have any effect.
Docusaurus does support node 18 so we should be OK

PR #187, opened 2022-09-13
[YUNIKORN-1295] document new annotations and how to use them

manirajv06:
Document changes to show the usage of new annotations with example.


wilfred-s (approved):
LGTM
one last typo left which I will fix on commit

PR #188, opened 2022-09-21
[YUNIKORN-1315] Document how to use MaxApplication enforcement feature

lixmgl:


wilfred-s (changes requested):
The limit object is linked to users and or groups. The queue configuration itself has a `maxApplications` entry. We should not update the part currently changed but the `Supported parameters for the queues` part. The list starts at line 133 of the file and should have max applications added. The queue example at line 177 needs to be updated to then show the updated entry.
Also make sure that the parent/child text is correct

wilfred-s:
I approved the workflow run. The update will run the workflow automatically from now on.
Sorry for the delay in the review, I am travelling at the moment.

lixmgl:
> The limit object is linked to users and or groups. The queue configuration itself has a `maxApplications` entry. We should not update the part currently changed but the `Supported parameters for the queues` part. The list starts at line 133 of the file and should have max applications added. The queue example at line 177 needs to be updated to then show the updated entry. Also make sure that the parent/child text is correct

Thanks @wilfred-s for reviewing the PR. I updated it according to your comments.
Have a safe trip!

docs/user_guide/queue_config.md
```diff
@@ -262,9 +272,10 @@ The star "*" is the wildcard character and matches all users or groups.
 Duplicate entries in the lists are ignored and do not cause a parsing error.
 Specifying a star beside other list elements is not allowed.
 
-_maxapplications_ is an unsigned integer value, larger than 1, which allows you to limit the number of running applications for the configured user or group.
+_maxapplications_ is an integer value, larger than 1, which allows you to limit the number of running applications for the configured user or group.
```
wilfred-s, line 275: unsigned was added on purpose as the parameter is defined as an `uint64` in the code. If you set a -1 an error unmarshalling the json will occur while a 0 should return a validation error. Probably better to keep the unsigned in the doc.

docs/user_guide/queue_config.md
```diff
@@ -262,9 +272,10 @@ The star "*" is the wildcard character and matches all users or groups.
...
 Specifying a zero maximum applications limit is not allowed as it would implicitly deny access.
-Denying access must be handled via the ACL entries.  
+Denying access must be handled via the ACL entries.
+_parent_ queue _maxapplications_ value must be larger than _parent_ queue _maxapplications_ value.
```
wilfred-s, line 278: This is not correct: the maxapplications for the user and or group must be smaller than the value set for the queue. However we do not enforce that at the moment. If we add a validation it will not be against the parent queue.

wilfred-s (changes requested):
Can we add this text between the description of sub queues and properties (line 163):
```
The _maxapplications_ property is an integer value, larger than 1, which allows you to limit the number of running applications for the queue. Specifying a zero for _maxapplications_ is not allowed as it would block all allocations for applications in the queue. The _maxapplications_ value for a _child_ queue must be smaller or equal to the value for the _parent_ queue.
```
We need to keep the queue and limit things completely separate.

docs/user_guide/queue_config.md
lixmgl: I see, the reason I updated here is @pbacsko commented https://github.com/apache/yunikorn-core/pull/429#discussion_r961046263
If you think we should keep unsigned type, i will revert and update PR 429 as well.

docs/user_guide/queue_config.md
lixmgl: Good to know. If we don't enforce, I will remove this line.

docs/user_guide/queue_config.md
```diff
@@ -133,6 +133,7 @@ Supported parameters for the queues:
 * name
 * parent
 * queues
+* maxApplications
```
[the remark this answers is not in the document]
lixmgl: updated to lowercase.

docs/user_guide/queue_config.md
lixmgl: Removed.

docs/user_guide/queue_config.md
wilfred-s: yes please do

wilfred-s:
please check the main [comment](https://github.com/apache/yunikorn-site/pull/188#pullrequestreview-1134942474) for the change around line 163

docs/user_guide/queue_config.md
lixmgl: reverted.

lixmgl:
> 



> 
sure, added.

wilfred-s (approved):
LGTM

PR #200, opened 2022-10-25
[YUNIKORN-1355] Generic example of GPU scheduling with Yunikorn

KatLantyss:
# Issue

Add a generic example of GPU scheduling with Yunikorn.
# Link

https://issues.apache.org/jira/browse/YUNIKORN-1355?filter=-1
# Solve

Test the GPU scheduling.
Update Yunikorn site.


yuchaoran2011 (commented):
Thanks @KatLantyss for your contribution. Is my understanding correct that YuniKorn as a scheduler treats GPUs, whether shared in a time-sliced manner or not, all the same? In other words, whether time slicing is enabled for GPUs shouldn't matter from a resource scheduling perspective, right?

wilfred-s:
> In other words, whether time slicing is enabled for GPUs shouldn't matter from a resource scheduling perspective, right?

That is correct this just brings the NVIDIA setup and the YuniKorn bits together.

docs/user_guide/workloads/run_nvidia.md
```diff
@@ -0,0 +1,263 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+To know how the GPU scheduling works, please refer to [**Time-Slicing GPUs in Kubernetes | Introduction**](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/gpu-sharing.html#introduction). This page covers ways to enable GPU scheduling in Yunikorn using [**NVIDIA GPU Operator**](https://catalog.ngc.nvidia.com/orgs/nvidia/containers/gpu-operator).
```
yuchaoran2011, line 28: Thanks @wilfred-s for the clarification. In that case, I think this part needs to be rephrased. The way it's written seems to suggest that enabling time slicing is a requirement for using GPUs, but in reality, it's an optional feature. We should make that clear. I've heard some production ML training use cases that in fact explicitly turned off time slicing since each work unit is heavy enough to fully use a single GPU. @KatLantyss

wilfred-s: @KatLantyss please check this comment

KatLantyss: @wilfred-s Sorry for the delay, I will change it recently, and thanks @yuchaoran2011 for the advice.

docs/user_guide/workloads/workload_overview.md
```diff
@@ -53,6 +53,7 @@ omitted as it will be set automatically on newly created pods.
 
 Examples of more advanced use cases can be found here:
 
+* [Run NVIDIA GPU Scheduling Jobs](run_nvidia)
```
yuchaoran2011, line 56: The title is `Run NVIDIA GPU Jobs` above. Here needs to be consistent

package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@docusaurus/core": "2.1.0",
     "@docusaurus/preset-classic": "2.1.0",
-    "@docusaurus/theme-search-algolia": "2.1.0",
+    "@docusaurus/theme-search-algolia": "^2.2.0",
```
yuchaoran2011, line 15: Can we leave the version update to another PR?

yuchaoran2011 (changes requested):

package.json
KatLantyss: @yuchaoran2011 Sure!

docs/user_guide/workloads/run_nvidia.md
```diff
@@ -0,0 +1,346 @@
...
+[Vector addition of 50000 elements]
+Copy input data from the host memory to the CUDA device
+CUDA kernel launch with 196 blocks of 256 threads
+Copy output data from the CUDA device to the host memory
+Test PASSED
+Done
+```
+
+---
+## Enable GPU Time-Slicing(Optional)
```
wilfred-s, line 112: nit: space between Slicing and the bracket.

wilfred-s:
Looks good now, one tiny nit
Checking with @yuchaoran2011 to see if anything is still left

package.json
yuchaoran2011: Looks like this change hasn't been reverted back

KatLantyss: @yuchaoran2011 Nailed it. Thanks!

yuchaoran2011 (approved):

PR #204, opened 2022-10-28
[YUNIKORN-1336] check git python install in release documentation

lowc1012:
Update the installing instruction for gitpython


wilfred-s (approved):
LGTM

PR #206, opened 2022-11-03

jagadeesanas2:
Verified the whole website and found multiple pages' URLs were broken. For example share feedback to the YuniKorn Community link on the website is broken.

They point to:
https://yunikorn.apache.org/community/get_involved/reporting_issues

While they should be pointing to the:
https://yunikorn.apache.org/community/reporting_issues/

### DIFF
--- a/src/pages/community/get_involved.md
+++ b/src/pages/community/get_involved.md
@@ -34,7 +34,7 @@ importance of this diversity and the value of "The Apache Way".  We welcome any 
 Please join us through one of the following channels:
 
 ### Contribute through github
-- Basic steps of contributing to project is explained here - [How to contribute](how_to_contribute)
+- Basic steps of contributing to project is explained here - [How to contribute](/community/how_to_contribute)
 
 ### Communication Channels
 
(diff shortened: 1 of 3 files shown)


wilfred-s:
Can you confirm that this is still broken for you?
I have tried all the URLs that you are changing and they show up and work correctly.
Maybe it is a browser version or type issue but would be good to find out

jagadeesanas2:
@wilfred-s I have created short video (YUNIKORN-1380_Recording.mp4) and attached over the [YUNIKORN-1380](https://issues.apache.org/jira/browse/YUNIKORN-1380)

jagadeesanas2:
@wilfred-s According to the [docusaurus doc](https://docusaurus.io/docs/markdown-features/links), it seems relative file paths are resolved against the current file's directory. Absolute file paths on the other hand are resolved relative to the content root, in our case _/community/_  That change only this PR proposed.

craigcondit:
If I open a debugger on the Network tab in Chrome Dev Tools, I can see that each new page clicked actually triggers a javascript load for each page, not a full page refresh. A browser refresh triggers the static content to be retrieved, and the yunikorn.apache.org site seems to add a trailing slash in that case.

Interestingly, if I disable all javascript via NoScript, I cannot duplicate the issue. This leads me to believe we need to update the trailingSlashes property to be true for docusaurus.

craigcondit (changes requested):
Let's address the root issue (the trailling slashes) instead.

jagadeesanas2:
Created new PR https://github.com/apache/yunikorn-site/pull/217

PR #212, opened 2022-11-17
[YUNIKORN-1406] Document new configuration system

craigcondit:
Updates the website with the configuration design documents and configuration reference.

JIRA: https://issues.apache.org/jira/browse/YUNIKORN-1406

### DIFF
--- a/docs/design/config_v2.md
+++ b/docs/design/config_v2.md
@@ -1,0 +1,345 @@
+---
+id: config_v2
+title: Configuration V2
+---
+
+<!--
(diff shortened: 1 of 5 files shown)


wilfred-s (approved):
LGTM with one remark: can we remove `package.json` from the commit? It is unneeded and we should keep the versions in sync. Not sure why the algolia search always updates independent of docusaurus.

craigcondit:
@wilfred-s yes, I'll remove package.json from the commit.

PR #217, opened 2022-11-26

jagadeesanas2:
Fix for trailingSlash issue.


craigcondit (approved):
+1, will merge shortly so we can test. It's possible we may need to revert if this makes things worse.

craigcondit (changes requested):
Build failed on master. Please test this locally to be sure it builds.

craigcondit (changes requested):
Syntax error during site generation.

jagadeesanas2:
Fixed syntax error and created new PR with the local test build.

PR #220, opened 2022-11-30
[YUNIKORN-1380] Multiple pages URL were broken for website

jagadeesanas2:
## Why are the changes needed?
Multiple pages URL were borken. For example share feedback to the YuniKorn Community link on the website is broken.

They point to:
https://yunikorn.apache.org/community/get_involved/reporting_issues

While they should be pointing to the:
https://yunikorn.apache.org/community/reporting_issues/

## What changes were proposed in this pull request?
Updated the trailingSlashes property to be true for docusaurus.

## How was this PR tested?
Tested with local-build 

image

PR #224, opened 2022-12-08

jagadeesanas2:
### Why are the changes needed?

Verified the whole website and found some broken pages' URLs even after trailingSlashes enabled. 

### What changes were proposed in this pull request?
This PR proposed absolute file paths on the other hand are resolved relative to the content root, in our case /community/

### How was this PR tested?
Tested with local-build

image


wilfred-s (approved):
LGTM: the docusaurus.config.js did not change looking at #220 as far as I can tell not sure why it is in the PR.
@craigcondit looks like al URLs are working

PR #227, opened 2022-12-12
YUNIKORN-1437: update the example REST response in scheduler page to include max runnin…

lixmgl:
https://issues.apache.org/jira/browse/YUNIKORN-1437


lixmgl:
@wilfred-s Please take a look, thanks.

PR #228, opened 2022-12-13
[YUNIKORN-1413] gang and CPU arch as features

wilfred-s:
Add gang scheduling and CPU architecture support to the features. Move max applications to next version to not lose it on the next release


craigcondit (approved):
+1 LGTM.

PR #231, opened 2022-12-13
[YUNIKORN-1418] Add scheduler state dump to Troubleshooting page

jagadeesanas2:
### Why are the changes needed?
An important tool to use during troubleshooting is the state dump. Collecting a state dump is not mentioned as part of the troubleshooting page.

### What changes were proposed in this pull request?
- Added ways to collect state dump.
- Updated note, that restarting the scheduler should be the last resort action 
- Verified troubleshooting page hyperlinks across other pages
- Updated sidebars.js

### How was this PR tested?
Tested with local-build


wilfred-s (approved):
LGTM

PR #232, opened 2022-12-16
[YUNIKORN-1475] Document priority support

craigcondit:
Document the priority support from YUNIKORN-1.

JIRA: https://issues.apache.org/jira/browse/YUNIKORN-1475


wilfred-s (approved):
LGTM
We need to make sure we do not forget to add the change for priority as a breaker to the release notes

craigcondit:
> LGTM We need to make sure we do not forget to add the change for priority as a breaker to the release notes

Indeed. I've added the `release-notes` tag to YUNIKORN-1.

PR #235, opened 2022-12-19
[YUNIKORN-1481] Broken links in zh-cn

KatLantyss:
fix link and file name.
截圖 2022-12-19 下午5 28 47


0yukali0 (approved):
I tried this commit and I could run the en-version and the zh-cn version.
LGTM.

PR #299, opened 2023-05-23
[YUNIKORN-1757] Placement Rules: fix complex example

mitdesai:
### What is this PR for?

The placement rules parent property takes 1 parent rule. The example shows that it can be an array. Which is incorrect. Updated the example to reflect the correct usage.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [X] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN-1757

Local testing, verified on locally deployed docker container.

### Questions:

### DIFF
--- a/docs/user_guide/placement_rules.md
+++ b/docs/user_guide/placement_rules.md
@@ -308,18 +308,18 @@ placementrules:
       groups:
         - dev*
     parent:
-      - name: tag
-        value: namespace
+      name: tag
+      value: namespace
   - name: tag
     value: namespace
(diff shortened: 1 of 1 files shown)


wilfred-s (approved):
LGTM, brings the example inline with what we already have in the test code

PR #321, opened 2023-06-28
[YUNIKORN-1831] Incorrect key name used for user info in pod annotati…

yzhangal:
…on in user guide example

### What is this PR for?
This PR is to fix the documentation about setting user info in annotation.

There are two issues:
1. The "username" word should be "user" which is implemented in code
2. There need to be quote around the keys "user", "groups"

### What type of PR is it?
* [ ] - Bug Fix
* [x ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* https://issues.apache.org/jira/browse/YUNIKORN-1831

Tested in locally built doc and content was tested in real cluster

### Questions:

### DIFF
--- a/docs/user_guide/usergroup_resolution.md
+++ b/docs/user_guide/usergroup_resolution.md
@@ -86,13 +86,15 @@ metadata:
   annotations:
     yunikorn.apache.org/user.info: "
     {
-      username: \"yunikorn\",
-      groups: [
+      \"user\": \"yunikorn\",
+      \"groups\": [
         \"developers\",
         \"devops\"
       ]
     }"
 ```
+
+(NOTE: for historical reason, "username" is used in label "yunikorn.apache.org/username", whereas, "user" is used in the annotation here)
 
 However, to enhance security, the following is enforced in the admission controller:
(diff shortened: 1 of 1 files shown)


docs/user_guide/usergroup_resolution.md
```diff
@@ -86,14 +86,16 @@ metadata:
...
-      groups: [
+      \"user\": \"yunikorn\",
+      \"groups\": [
         \"developers\",
         \"devops\"
       ]
     }"
 ```
 
+(NOTE: for historical reason, "username" is used in label "yunikorn.apache.org/username", whereas, "user" is used in the annotation here)
```
craigcondit, line 97: I think we should leave the note out as it's completely obsolete. It's likely to cause more confusion than not. The old label still exists but has completely different meaning and semantics.

craigcondit (changes requested):
Looks ok, other than added note (which should be removed).

craigcondit (approved):
+1 LGTM.

PR #325, opened 2023-07-28
[YUNIKORN-1884] fix allow-preemption

FrankYang0529:
### What is this PR for?

The `allow-preemption` can be set on the Pod or PriorityClass object.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [X] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-1884

### Questions:


craigcondit (changes requested):
Looks good, but remove the changes to the versioned_docs -- we don't modify those once we have shipped a release.

FrankYang0529:
> Looks good, but remove the changes to the versioned_docs -- we don't modify those once we have shipped a release.

Reverted the change in versioned_docs. Thanks!

PR #336, opened 2023-09-14
[YUNIKORN-1982] Enable manual triggering for GitHub CI

doupache:
### What is this PR for?
Allow running manually from the github ui. 
Everyone can manually run the CI on their fork.

github docs 
1. https://docs.github.com/en/actions/using-workflows/manually-running-a-workflow
2. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

This will not affect **ANY** github action we are running.
### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
 [YUNIKORN-1982](https://issues.apache.org/jira/browse/YUNIKORN-1982)

### DIFF
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -19,7 +19,6 @@ on:
   push:
     branches:
       - master
-  workflow_dispatch: {}
 
 jobs:
   build:
--- a/.github/workflows/auto-publish.yml
+++ b/.github/workflows/auto-publish.yml
@@ -19,6 +19,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch: {}
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/auto-publish.yml
```diff
@@ -19,6 +19,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch: {}
```
wilfred-s, line 22: We do not want the push flow to be run manually.
If something is that broken that a publishing action to the production site needs to be run manually then we need to run it fully manually.

wilfred-s (changes requested):

doupache:
@wilfred-s  
We only manual trigger pre-commit hook. 
Please Take a look. 🙏

pbacsko:
+1 LGTM

craigcondit (approved):
+1 LGTM.

PR #349, opened 2023-10-06
[YUNIKORN-2014] update /ws/v1/config response

FrankYang0529:
### What is this PR for?

Update `/ws/v1/config` response.

### What type of PR is it?
* [X] - Documentation

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2014

### DIFF
--- a/docs/api/scheduler.md
+++ b/docs/api/scheduler.md
@@ -1578,20 +1578,72 @@ Endpoint to retrieve the current scheduler configuration
 
 **Code** : `200 OK`
 
-**Content example**
+**Content example (with `Accept: application/json` header)**
+
+```json
+{
+    "Partitions": [
+        {
+            "Name": "default",
+            "Queues": [
+                {
+                    "Name": "root",
+                    "Parent": true,
+                    "Resources": {},
+                    "SubmitACL": "*",
+                    "ChildTemplate": {
+                        "Resources": {}
+                    }
+                }
+            ],
+            "PlacementRules": [
+                {
+                    "Name": "tag",
+                    "Create": true,
+                    "Filter": {
+                        "Type": ""
+                    },
+                    "Value": "namespace"
+                }
+            ],
+            "Preemption": {
+                "Enabled": false
+            },
+            "NodeSortPolicy": {
+                "Type": ""
+            }
+        }
+    ],
+    "Checksum": "FD5D3726DF0F02416E02F3919D78F61B15D14425A34142D93B24C137ED056946",
+    "Extra": {
+        "event.trackingEnabled": "false",
+        "log.core.scheduler.level": "info",
+        "log.core.security.level": "info",
+        "log.level": "debug"
+    }
+}
+```
+
+**Content example (without `Accept: application/json` header)**
 
 ```yaml
 partitions:
-- name: default
-  queues:
-  - name: root
-    parent: true
-    submitacl: '*'
-  placementrules:
-  - name: tag
-    create: true
-    value: namespace
-checksum: D75996C07D5167F41B33E27CCFAEF1D5C55BE3C00EE6526A7ABDF8435DB4078E
+    - name: default
+        queues:
+            - name: root
+            parent: true
+            submitacl: "*"
+        placementrules:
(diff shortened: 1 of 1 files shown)


manirajv06 (commented):
looks good.

My understanding is, https://github.com/apache/yunikorn-core/pull/652 changes won't break any existing REST API consumers on parsing the response. Can you confirm this?

FrankYang0529:
> looks good.
> 
> My understanding is, [apache/yunikorn-core#652](https://github.com/apache/yunikorn-core/pull/652) changes won't break any existing REST API consumers on parsing the response. Can you confirm this?

Yeah, there is a bug. Created a PR https://github.com/apache/yunikorn-core/pull/672 to fix it.

FrankYang0529:
> > > looks good.
> > > My understanding is, [apache/yunikorn-core#652](https://github.com/apache/yunikorn-core/pull/652) changes won't break any existing REST API consumers on parsing the response. Can you confirm this?
> > 
> > 
> > Yeah, there is a bug. Created a PR [apache/yunikorn-core#672](https://github.com/apache/yunikorn-core/pull/672) to fix it.
> 
> I don't think so. Otherwise, below assertions should have failed
> 
> https://github.com/apache/yunikorn-core/blob/master/pkg/webservice/handlers_test.go#L510 https://github.com/apache/yunikorn-core/blob/master/pkg/webservice/handlers_test.go#L542
> 
> in the first place. _go embedded field is implicitly inlined_ (you can refer [this](https://pkg.go.dev/github.com/go-json-experiment/json#:~:text=A%20Go%20embedded%20field%20is,unnamed%20pointer%20to%20such%20types) ) . Even now, https://github.com/apache/yunikorn-core/pull/672/files has taken care only for yaml, not for json. Hence, let's leave as is.

I think the test case can work is because we use the same struct. For main branch version, both don't have inline tag. For PR branch version, both have inline tag.

May you check following issues and go playground, so we can see the different part.

* https://github.com/go-yaml/yaml/issues/99
* https://github.com/go-yaml/yaml/issues/63
* https://go.dev/play/p/0Qa5Wcb94q9

manirajv06:
It seems unmarshal behaviour is different for yaml and json. To be defensive, let's keep the inline tag.

> For main branch version, both don't have inline tag. For PR branch version, both have inline tag.

I can see only for yaml, not for json. Should we do something like this?

`*configs.SchedulerConfig `json:",inline" yaml:",inline"``

FrankYang0529:
> I can see only for yaml, not for json. Should we do something like this?
> 
> `*configs.SchedulerConfig `json:",inline" yaml:",inline"``

Add json inline tag in https://go.dev/play/p/pwH25J2iuOc and the result is same as previous one https://go.dev/play/p/0Qa5Wcb94q9.

I think we probably don't need to add inline tag for json.

FrankYang0529:
> @FrankYang0529 Will approve the pr and merge the code. Just for our better understanding, It would be great if you can try hitting the url in your local setup after yk installation and share your observations.

Sounds good. Thank you. I will leave a comment in both Jira ticket and this PR for better tracking.

FrankYang0529:
YAML response **without** `inline` tag.
```yaml
schedulerconfig:
    partitions:
        - name: default
          queues:
              - name: root
                parent: true
                submitacl: "*"
          placementrules:
              - name: tag
                create: true
                value: namespace
    checksum: FD5D3726DF0F02416E02F3919D78F61B15D14425A34142D93B24C137ED056946
extra:
    event.trackingEnabled: "false"
    log.core.scheduler.level: info
    log.core.security.level: info
    log.level: debug
```

YAML response **with** `inline` tag.
```yaml
partitions:
    - name: default
        queues:
            - name: root
            parent: true
            submitacl: "*"
        placementrules:
            - name: tag
            create: true
            value: namespace
checksum: FD5D3726DF0F02416E02F3919D78F61B15D14425A34142D93B24C137ED056946
extra:
    event.trackingEnabled: "false"
    log.core.scheduler.level: info
    log.core.security.level: info
    log.level: debug
```

PR #359, opened 2023-10-23
[YUNIKORN-2035] Update queue api example in scheduler api document

Haser0305:
### What is this PR for?
Adding tag `details` to list all the fields and explanations. I believe this can make it easier for new users to understand the API. Moreover, it can mark the field is `omiempty` or not. For example,  the `parent: "",` should not be contained in the response json in the current api version. 

If this format is too abrupt, consider using a list to display the fields that are `omiempty`.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-2035

image


wilfred-s:
I think we can leave out the mention of the omit empty for each field and make it a generic mention. The fields that are not marked with omit on empty always have a value. They cannot be empty.
There is only one exception to this and that are the boolean fields. Boolean fields are the only ones not marked with omit on empty. The "empty" value for the boolean is false. Having the boolean value always there was considered the better solution from a supportability perspective.

The types you are exposing are the go specific types and not the json types. Would it not be a better idea to expose the json schema for the objects instead of the go types?

Haser0305:
> I think we can leave out the mention of the omit empty for each field and make it a generic mention. The fields that are not marked with omit on empty always have a value. They cannot be empty. There is only one exception to this and that are the boolean fields. Boolean fields are the only ones not marked with omit on empty. The "empty" value for the boolean is false. Having the boolean value always there was considered the better solution from a supportability perspective.

Understand. How about replacing `omitted if empty` with `allow empty`? 
This could more clearly express whether the field can be empty or not. And together with the additional explanation outside the table, it can be understood that the json response may omit this field due to empty values. 

> The types you are exposing are the go specific types and not the json types. Would it not be a better idea to expose the json schema for the objects instead of the go types?

I completely agree with your point. I will make the changes. 

Thank you for the feedback.

docs/api/scheduler.md
```diff
@@ -31,6 +31,9 @@ Many of these APIs return collections of resources. Internally, all resources ar
 are returned in units of bytes while resources of type `vcore` are returned in units of millicores
 (thousands of a core). All other resource types have no specific unit assigned.
 
+Additionally, it should be note that in the JSON response bodies, some fields may be omitted if their values are empty. Specifically, if a field is not present in the response, it can be assumed that the value is empty/null.
```
wilfred-s, line 34: We can use the following markup type to show this in a nicer way:
```
:::note PUT TEXT HERE :::
```
That would allow the removal of: "Additionally, it should be note that"
See https://docusaurus.io/docs/markdown-features/admonitions

wilfred-s (changes requested):

PR #376, opened 2023-12-01
[YUNIKORN-2226] Add updatePreemptionInfo to the mutation webhook doc

doupache:
### What is this PR for?
Add updatePreemptionInfo to the mutation webhook doc

### What type of PR is it?
* [x] - Improvement

### What is the Jira issue?
[YUNIKORN-2226](https://issues.apache.org/jira/browse/YUNIKORN-2226)


doupache:
cc @craigcondit

pbacsko:
@doupache this is archived design. I don't think we need to update this. The admission controller part looks valid, but it's archived for a reason.

pbacsko (changes requested):
Suggest dropping this PR. Don't touch archived docs, they might be no longer valid.

PR #379, opened 2023-12-06
[YUNIKORN-2217] Add Missing REST Endpoint description for batch event in documentation

Haser0305:
### What is this PR for?
Adds documentation for the /ws/v1/events/batch API endpoint. 

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2217

### DIFF
--- a/docs/api/scheduler.md
+++ b/docs/api/scheduler.md
@@ -1895,3 +1895,68 @@ None
 ### Error response
 
 **Code**: `400 Bad Request`
+
+
+## Batch Events
+
+Endpoint is used to retrieve a batch of event records.
+
+**URL**: `/ws/v1/events/batch`
+
+**METHOD** : `GET`
+
+**Auth required** : NO
+
+### Success response
+
+**Code**: `200 OK`
+
+**Content examples**
+
+```json
+{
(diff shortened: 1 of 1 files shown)


wilfred-s (approved):
LGTM, waiting for a review from @pbacsko before committing to make sure we have covered the latest changes

pbacsko (changes requested):
Also document the URL query parameters:
* `count`: maximum number of events in the response
* `start`: the start ID of the events

eg. `localhost:9080/ws/v1/events/batch?start=200&count=1000`

If the ID is not in the ring buffer (too low or too high), then the response contains the current lowest and highest ID in the response with `EventRecords` being empty.

See
https://github.com/apache/yunikorn-core/blob/681a4cbf4b9929e579674e44e6309acf0e728e29/pkg/events/event_system.go#L57-L61

https://github.com/apache/yunikorn-core/blob/681a4cbf4b9929e579674e44e6309acf0e728e29/pkg/webservice/handlers.go#L1055-L1082

pbacsko (approved):
+1

PR #387, opened 2023-12-28
[YUNIKORN-2256] Fix link error occurs when clicking link on page with trailing slash in the URL

brandboat:
### What is this PR for?
I've observed that many links in the Chinese documentation can't be converted to markdown relative file paths because some documents are not yet translated into Chinese. The task of changing all links to markdown file paths is currently not possible.
In the pull request:
1. setting trailingSlash to true
2. fixing any links that become invalid due to this setting.

### What type of PR is it?
* [x] - Bug Fix

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2256

covered by `yarn build` in docker images, in docusaurus, `yarn build` will check if there is any broken links, if there is any, error occurs.

N/A

### Questions:
N/A


wilfred-s:
We seem to be using three distinct types of URL references:

1. `../path/file`
2. `path/file`
3. `path/file.md`

In the PR it looks like you did not touch the 3rd version and only changed the 2nd version into the 1st version. Does that mean the 3rd version worked correctly?
Would be good to standardise on one option. That seems to leave us with an option of choosing between 1st and 3rd. I don't think it matters much which one we choose. Making a choice and enforcing it is more important.
We can then even enforce it via a build check. Making sure all links comply and prevent breakages.

brandboat:
> In the PR it looks like you did not touch the 3rd version and only changed the 2nd version into the 1st version. Does that mean the 3rd version worked correctly?

Yes, it works fine. We can use both url path or file path in markdown links. (https://docusaurus.io/docs/2.x/markdown-features/links)

> Would be good to standardise on one option. That seems to leave us with an option of choosing between 1st and 3rd. I don't think it matters much which one we choose. Making a choice and enforcing it is more important.
We can then even enforce it via a build check. Making sure all links comply and prevent breakages.

There are two reason some links still use 1st version and 2nd version, and can't be switched to 3rd version.
1. Some documents are not translated to Chinese yet, so the 3rd won't work since the corresponding markdown files not exist, and we use 1st and 2nd to let markdown links direct to English document.
2. The markdown links in files under `src/pages/community` can't be switched to 3rd version, not sure why this is not allowed. (Looks like links in 3rd version only works in [docs](https://docusaurus.io/docs/2.x/docs-introduction), not [pages](https://docusaurus.io/docs/2.x/creating-pages))

If we want to standardize the markdown links format, I believe we can only use the 1st version or the 2nd version, which means the URL path. Perhaps we can open another JIRA to handle this ? I think this will require extensive changes, considering that the majority of links currently use the 3rd version.

wilfred-s:
Assigned and added a lot of reviewers. If we can get this working before we add 1.5 then I think we should do it.
Broken link detections seems to be working @brandboat 👍

chia7712:
@brandboat Please rebase code :)

brandboat:
After discussing offline with @chia7712 , I've realized that the proposed solution in this pull request, which adds `../` to each link, is not intuitive and could confuse developers. The reason for adding `../` in most cases is due to the missing Chinese documentation. Therefore, I plan to address this issue in a new pull request after resolving https://issues.apache.org/jira/browse/YUNIKORN-2411.

PR #389, opened 2024-01-07
[YUNIKORN-2314] Update partition limit documentation

brandboat:
### What is this PR for?
Now partition limits works the same as root queue limits, while if partition limits and root queue limits both exist and they are not equivalent, configvalidator will throw error. We need to mention this behavior in document.

### What type of PR is it?
* [x] - Documentation

### Todos
N/A

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2314

N/A

N/A

### Questions:
N/A


chia7712 (approved):
LGTM

PR #393, opened 2024-01-26
[YUNIKORN-2267] add docs of using BinaryData to configure scheduler

FrankYang0529:
### What is this PR for?

Add `binaryData` config example.

### What type of PR is it?
* [X] - Documentation

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2267

### DIFF
--- a/docs/user_guide/service_config.md
+++ b/docs/user_guide/service_config.md
@@ -977,6 +977,40 @@ Example:
 admissionController.accessControl.externalGroups: "^sales$,^marketing$,^admin-"
 ```
 
+### Using compressed values
+
+The data in ConfigMap cannot exceed 1 MiB. YuniKorn supports the gzip algorithm to decompress data in the `binaryData` field.
+If a key ends with `.gz`. YuniKorn will treat the value as gzip-compressed data and decompress it automatically. The base64 encoding is automatically.
+If a value is set in both the `data` and `binaryData` sections, the value in the `binaryData` section will be used.
+
+Example:
+
+Users can run the command to get the value.
+
+```bash
+echo "
+partitions:
+  - name: default
+    queues:
+      - name: root
+        submitacl: '*'
+        parent: true
+        queues:
+          - name: parent
+            submitacl: '*'" | gzip | base64
+```
+
+Set the result in the `binaryData` field.
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: yunikorn-configs
+binaryData:
+  queues.yaml.gz: "H4sIAMyHs2UAA2WMSQ6AIBAE77yibyQmfoDfoI4JCYvCzP/FjWDsY3Wl1GYzO3YpFqOAEdEGMlhoteK5EmAXErrec6+RU+IHAUWm4NjO3kAPuuHapsgGnIUa/Ob65K13xy98AFwE9HmuAAAA"
+```
+
 ### Deprecated settings
 
 #### service.operatorPlugins


docs/user_guide/service_config.md
```diff
@@ -977,6 +977,37 @@ Example:
 admissionController.accessControl.externalGroups: "^sales$,^marketing$,^admin-"
 ```
 
+### Compresion setting
+
+The data in ConfigMap cannot exceed 1 MiB. YuniKorn supports Gzip algorithm to decompress data in the `binaryData` field.
```
craigcondit, line 982: Nit: "YuniKorn supports **_the gzip_** algorithm" instead of "YuniKorn supports Gzip algorithm" (gzip is lowercased, and should have the article "the" preceding it).

docs/user_guide/service_config.md
```diff
@@ -977,6 +977,37 @@ Example:
 admissionController.accessControl.externalGroups: "^sales$,^marketing$,^admin-"
 ```
 
+### Compresion setting
```
craigcondit, line 980: Nit: Compresion setting -> Using compressed values

docs/user_guide/service_config.md
```diff
@@ -977,6 +977,37 @@ Example:
...
+
+The data in ConfigMap cannot exceed 1 MiB. YuniKorn supports Gzip algorithm to decompress data in the `binaryData` field.
+If a key ends with `.gz`, YuniKorn decodes the value by base64 and decompress it by Gzip algorithm.
```
craigcondit, line 983: How about: "If a key ends with `.gz`. YuniKorn will treat the value as gzip-compressed data and decompress it automatically. The base64 encoding is automatically "

docs/user_guide/service_config.md
```diff
@@ -977,6 +977,37 @@ Example:
...
+If both `queues.yaml` setting is set in `data` field and `queues.yaml.gz` setting is set in `binaryData` field, the value in `binaryData` field will be used.
```
craigcondit, line 984: Instead: If a value is set in both the `data` and `binaryData` sections, the value in the `binaryData` section will be used.

craigcondit (changes requested):
Minor changes, otherwise looks good.

chia7712 (commented):
@FrankYang0529 nice docs. some trivial comments below.

docs/user_guide/service_config.md
craigcondit: We don't want to log a warning because it's not uncommon that a default queues.yaml is provided via Helm in the yunikorn-defaults Condit so that is then overridden by a binaryData version in yunikorn-configs. It's not possible to remove a config value so having both would create an "unfixable" warning.

chia7712: > it's not uncommon that a default queues.yaml is provided via Helm in the yunikorn-defaults configmap that is then overridden by a binaryData version in yunikorn-configs. 

make sense to me. thanks

FrankYang0529:
Hi @chia7712 and @craigcondit, thanks for reviewing. I resolved all comments. Could you help to take a look again? Thanks.

docs/user_guide/service_config.md
```diff
@@ -243,7 +243,7 @@ Sets the tolerations for the YuniKorn scheduler pod.
 
 Default: `[]`
 
-Example:
```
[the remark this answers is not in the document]
FrankYang0529: I disable trim trailing whitespace in my IDE, so we don't have unrelated change in this PR. Thanks!

chia7712 (approved):
LGTM

PR #406, opened 2024-02-29
[YUNIKORN-2444] Create user guide for the event system

pbacsko:
### What is this PR for?
User guide to the event system & history tracking.  

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2444

### Questions:

### DIFF
--- a/docs/user_guide/event_system.md
+++ b/docs/user_guide/event_system.md
@@ -1,0 +1,147 @@
+---
+id: event_system
+title: event_system
+---
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Overview
+
+The Yunikorn scheduler core generates well-defined events at various points during the execution. Since an event has a fixed structure, it is an ideal input for automated tools for further processing. Although Yunikorn has extensive logging, the output is textual and needs to be parsed. Not only is it error-prone, but also nothing guarantees that it never changes over time: it might be completely removed in a future release or the log level can change. 
+
+The event object is defined in the scheduler interface repository along with the constants that describe a certain event.
+
+Certain types of events which are related to pods and nodes are also sent to Kubernetes itself. This allows the users to identify certain problems that can arise from misconfiguration, such as insufficient queue or user quota. If a pod is pending because the quota is exceeded, it will be updated with an event about why it cannot be scheduled. Previously this had to be inferred from the logs or state dump.
+
+Various properties of the event subsystem are configurable. Changes are picked up and applied immediately.
+
+# Event types
+
+Events always describe a specific object type. There are four types of events: application, queue, node and user/group.
+When it is appropriate, the event will be sent to Kubernetes using the K8s event API. Since the API requires a "regarding" object (ie. which K8s object the event is about), we only send node and pod events.
+
+The following table summarizes when they are generated.
+
+| Type | Event | Notes |
+|--|--|--|
+| Application | A new application is created ||
+| Application | Application state transition ||
+| Application | Successful allocation ||
+| Application | An ask is added to the application ||
+| Application | Application removed ||
+| Application | Application rejected ||
+| Application | Ask removed ||
+| Application | Allocation removed ||
+| Application | Placeholder timeout ||
+| Application | Allocation replacement (gang scheduling) ||
+| Application | Application is not runnable due to maxApplications | Queue or user maxApps limit |
+| Application | Application has become runnable | Only sent if limit was hit before |
+| Application | Ask is not schedulable due to quota limitations | Sent to Kubernetes to the respective pod |
+| Application | Ask has become schedulable (there is quota now) | Sent to Kubernetes to the respective pod<br/> Only sent if quota was not available before |
+| Application | Ask is not schedulable due to predicate error * | Sent to Kubernetes to the respective pod<br/> Rate limited: sent once in every 15 seconds |
+| Application | Real allocation is larger than the placeholder | Sent to Kubernetes to the respective pod |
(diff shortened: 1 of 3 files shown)


docs/user_guide/event_system.md
```diff
@@ -0,0 +1,147 @@
...
+
+The streaming endpoint is available at `/ws/v1/events/stream`. Unlike the batch interface, which closes the connection after the query, streaming keeps the HTTP connection open. There is no idle timeout, so as long as the connection is stable, it is never closed. New events are sent immediately to active clients. The URL parameter "count" is also accepted; it tells Yunikorn how many recent events we want to fetch that were generated before the connection.
+
+Since this approach uses more resources inside Yunikorn, the number of streaming connections is limited, but the limits are configurable. See [streaming](../api/scheduler.md#event-stream) for details. Existing connections are not closed after configuration change.
+
+The active streaming connections are also shown in the state dump.
+
+# Event structure
+
+A Yunikorn event has the following fields:
```
chia7712, line 113: Does it include the change of YUNIKORN-2454?

pbacsko: That's separate from `si.EventRecord`.

chia7712: My point was: should we remind users the record introduced by YUNIKORN-2454? The record is different from `si.EventRecord` so user should "parse" it by another way.

docs/user_guide/event_system.md
```diff
@@ -0,0 +1,147 @@
...
+
+| Name | Description | Always set |
+|--|--|--|
+| ObjectID | Identifies the object what the event is about. | Yes |
+| ReferenceID | Identifies a secondary object which is related to the first one. It can be an allocation ID, queue, etc. | No |
+| Message| Textual details. Relevant if it is sent to Kubernetes. | No |
+| Type | The type of the ObjectID: application, queue, node, user/group. | Yes |
+| EventChangeType | The nature of the change: adding, setting, removing or none. | Yes |
+| EventChangeDetail | Further details about the change itself. | Yes |
```
chia7712, line 122: IIRC, `EventChangeDetail` field can be omitted by the restful API if the value is zero. Is that conflict with "Always set"?

docs/user_guide/event_system.md
pbacsko: I think it's only relevant in the REST API docs. However, we still need changes on that front (eg. at least the starting ID need to be included). So we can fix that later.

docs/user_guide/event_system.md
pbacsko: Yeah it's a bit of semantics, since it cannot be nil and zero is actually a valid value. Same goes for EventChangeType.

chia7712 (approved):
LGTM

docs/user_guide/event_system.md
```diff
@@ -0,0 +1,147 @@
+---
+id: event_system
+title: event_system
```
chenyulin0719, line 3: The title should be "Event System". 

image

docs/user_guide/event_system.md
```diff
@@ -0,0 +1,147 @@
...
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Overview
```
chenyulin0719, line 25: Could you add 1 more heading level? Change all '#' to '##', '##' to '###'. 

The current doc index is:
image


I think this would be better:

image

docs/user_guide/event_system.md
```diff
@@ -0,0 +1,147 @@
...
+| User/group | Resource usage increased for a group | Only sent if group limit is active |
+| User/group | Group tracking got associated with a user| Only sent if group limit is active |
+| User/group | Group tracking removed from a user| Only sent if group limit is active |
+
+`*` a predicate is a plugin located in the default scheduler. They're responsible for functionality like node selectors, affinity, anti-affinity, etc. Yunikorn runs the predicates for every allocation to see if it fits the candidate node.
+
+# In-memory storage of historical events
+
+Yunikorn stores a previously generated event objects in a ring buffer. Once it gets full, oldest elements are overwritten. The default size of the buffer is 100000.
+On a busy cluster this might not be enough. The size can be changed in the configmap by using the key `event.ringBufferCapacity`.
```
chenyulin0719, line 85: I didn't see below configs in  [YuniKorn Configuration](https://yunikorn.apache.org/docs/next/user_guide/service_config#yunikorn-configuration) page

- event.ringBufferCapacity
- event.trackingEnabled

Is it missing?
Is '[service.eventChannelCapacity](https://yunikorn.apache.org/docs/next/user_guide/service_config#serviceeventchannelcapacity)' equal to 'event.ringBufferCapacity'?

chenyulin0719 (changes requested):
2 change requests and 1 question.

docs/user_guide/event_system.md
pbacsko: Actually, four properties are missing.

`event.requestCapacity` is completely ignored now, it should be fixed.

docs/user_guide/service_config.md
```diff
@@ -651,6 +651,49 @@ Example:
...
+#### event.maxStreamsPerHost
+Sets the maximum number of event stream connections from a given host.
+
+Default: `15`
+
+Example:
+```yaml
+event.maxStreamsPerHost: "5"
+```
+
```
chenyulin0719, line 696: Hi @pbacsko, 
One minor question, the event.requestCapacity is configurable but it never been used in core(Except unit test). Is it designed for future changes?
- https://github.com/apache/yunikorn-core/blob/b44cc5ddd2c4bdb19e7a03cb02aa1eff4a8da69b/pkg/common/configs/configs.go#L37

chenyulin0719 (approved):
+1 LGTM, 

I'm not sure what is the original intention behind supporting `event.requestCapacity`.
Please create a new Jira to either remove or implement the logic for `event.requestCapacity`. Regards.

docs/user_guide/event_system.md
chenyulin0719: @chia7712 The missing configs are fixed in the latest commit, so I think https://issues.apache.org/jira/browse/YUNIKORN-2455 could be closed.

chia7712: > The missing configs (4 of 5) are fixed in the latest commit, so I think https://issues.apache.org/jira/browse/YUNIKORN-2455 could be closed.

done

pbacsko:
I created https://issues.apache.org/jira/browse/YUNIKORN-2512 to address event.requestCapacity. After that is done, I'll update the docs again.

PR #423, opened 2024-04-21
[YUNIKORN-2471] Documentation update for max 0 queue resources.

tommyfufu:
### What is this PR for?
A few sentences describing the overall goals of the pull request's commits.
First time? Check out the contributing guide - http://yunikorn.apache.org/community/how_to_contribute   

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
* Open an issue on Jira https://issues.apache.org/jira/browse/YUNIKORN/
* Put link here, and add [YUNIKORN-*Jira number*] in PR title, eg. `[YUNIKORN-2] Gang scheduling interface parameters`

### Questions:

### DIFF
--- a/docs/user_guide/queue_config.md
+++ b/docs/user_guide/queue_config.md
@@ -395,6 +395,7 @@ There is no guaranteed resource setting for the root queue.
 
 Maximum resources when configured place a hard limit on the size of all allocations that can be assigned to a queue at any point in time.
 A maximum resource can be set to 0 which makes the resource not available to the queue.
+However, setting all maximum resources to zero will be ignored and logged for debugging.
 Guaranteed resources are used in calculating the share of the queue and during allocation. 
 It is used as one of the inputs for deciding which queue to give the allocation to.
 Preemption uses the _guaranteed_ resource of a queue as a base which a queue cannot go below.


docs/user_guide/queue_config.md
```diff
@@ -395,6 +395,7 @@ There is no guaranteed resource setting for the root queue.
 
 Maximum resources when configured place a hard limit on the size of all allocations that can be assigned to a queue at any point in time.
 A maximum resource can be set to 0 which makes the resource not available to the queue.
```
wilfred-s, line 397: change from:
> A maximum resource can be set to 0 which makes the resource not available to the queue.

to:
A single resource type within the maximum resource can be set to 0 which makes that type not available to the queue.

docs/user_guide/queue_config.md
```diff
@@ -395,6 +395,7 @@ There is no guaranteed resource setting for the root queue.
...
+However, setting all maximum resources to zero will be ignored and logged for debugging.
```
wilfred-s, line 398: Change from:
> However, setting all maximum resources to zero will be ignored and logged for debugging.

to:
However, a maximum resource with all resource types set to 0 will be ignored, not set on the queue, and logged for debugging.

wilfred-s (changes requested):

docs/user_guide/queue_config.md
chenyulin0719: I just noticed that setting all the maximum resource to 0 does not simply get ignored. 
If a maximum resource setting already exists in the queue, the previous setting will be cleared.

This behavior is not documented too. 

https://github.com/apache/yunikorn-core/blob/ff654b2dd9038c1b026d99e553f79cd32e3d4de7/pkg/scheduler/objects/queue.go#L407-L408

wilfred-s: ah yes that is correct, that was introduced via YUNIKORN-2169 to allow removing a quota. if the queue is configured and no longer has a quota.
Good catch. We need to add that behaviour as it was a fix requested but never documented.

PR #429, opened 2024-05-14
[YUNIKORN-2613] Release notes for 1.5.1

pbacsko:
### What is this PR for?
Add release annoucement page for 1.5.1 

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2613

### Questions:
* [ ] - The licenses files need update.
* [ ] - There is breaking changes for older versions.
* [ ] - It needs documentation.

PR #431, opened 2024-05-17
[YUNIKORN-2614] Update website for 1.5.1

pbacsko:
### What is this PR for?
Update website for release 1.5.1

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2614

### Questions:


pbacsko:
ok, need to merge #429 first

craigcondit (commented):
I wonder if we shouldn't skip the versioned_docs update and do the release announcement only. There are no documentation updates between 1.5.0 and 1.5.1, so we just introduce another version that needs to be built for every change moving forward.

wilfred-s:
I think we need to do it the other way around: remove 1.5.0 and use 1.5.1 as the release. We do not want people to use 1.5.0 as the deadlocks are serious enough to make it a bad release.
Maybe also remove 1.5.0 from the downloads and replace with 1.5.1

wilfred-s (approved):
LGTM
We need to sort out what we do with patch releases and documentation and add it to the release procedure as part of the next release. Making ad-hoc decisions is not the right approach

PR #452, opened 2024-07-08
Appoint specific version when installing yunikorn

Paul-Lyu:
### What is this PR for?
Appoint specific version when installing yunikorn  
In get started doc, image tags are latest which is not available on docker hub.

### What type of PR is it?
* [ ] - Bug Fix
* [X] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?

https://issues.apache.org/jira/browse/YUNIKORN-2696

### Questions:

### DIFF
--- a/docs/get_started/get_started.md
+++ b/docs/get_started/get_started.md
@@ -50,6 +50,27 @@ admission controller delegating all scheduling to YuniKorn. Because this mode is
 
 If you are unsure which deployment mode you should use, refer to our [side-by-side comparison](user_guide/deployment_modes).
  
+If you want to use specific Yunikorn version, you can update helm chart via helm upgrade.
+```shell script
+helm upgrade -f custom.yml  -install yunikorn yunikorn/yunikorn -n yunikorn --create-namespace
+```
+
+The custom.yml is
+```yml
+image:
+  tag: scheduler-1.5.1
+
+admissionController:
+  image:
+    tag: admission-1.5.1
+
+web:
+  image:
+    tag: web-1.5.1
+```
+
+The `latest` image tags which is not available on [docker hub](https://hub.docker.com/r/apache/yunikorn/tags).
+
 Further configuration options for installing YuniKorn via Helm are available in the [YuniKorn Helm hub page](https://hub.helm.sh/charts/yunikorn/yunikorn).
 
 If you don't want to use helm charts, you can find our step-by-step


chia7712 (approved):
LGTM

craigcondit (changes requested):
I don't think we should hardcode a specific version here. This will get out-of-date quickly. Instead, reference the tags as "scheduler-{version}" and point users to DockerHub for available tags.

Also, drop reference to 'latest'. 'docker hub' should be 'DockerHub', and 'Yunikorn' should be 'YuniKorn'.

craigcondit (changes requested):
See previous comment.

chia7712:
> I don't think we should hardcode a specific version here. This will get out-of-date quickly. Instead, reference the tags as "scheduler-{version}" and point users to DockerHub for available tags.

agree. This PR can align the example of https://github.com/apache/yunikorn-site/blob/master/docs/user_guide/service_config.md?plain=1#L39 

> 'Yunikorn' should be 'YuniKorn'.

There are many such typo in code base. Maybe we should fix it. at least yunikorn-site repo

ryankert01:
> I don't think we should hardcode a specific version here. This will get out-of-date quickly. Instead, reference the tags as "scheduler-{version}" and point users to DockerHub for available tags.

> agree. This PR can align the example of https://github.com/apache/yunikorn-site/blob/master/docs/user_guide/service_config.md?plain=1#L39

I agree; we can align this method with all versioned examples in the docs.

wilfred-s (approved):
LGTM, I think all requested changes are done

chia7712 (commented):
@Paul-Lyu thanks for updates. Could you please update https://github.com/apache/yunikorn-site/blob/master/docs/user_guide/service_config.md#container-images too?

chenyulin0719:
Sorry for the late review.
Instead of providing a custom.yml, I would prefer to use a single command.

EX:
```
VERSION=1.5.2
helm install yunikorn yunikorn/yunikorn --namespace yunikorn \
  --set image.tag=scheduler-${VERSION} \
  --set pluginImage.tag=scheduler-plugin-${VERSION} 
  --set admissionController.image.tag=admission-${VERSION} \
  --set web.image.tag=web-${VERSION}
```

The better solution is to extract variable `version` in helm's values.yml, so we could use:

```
VERSION=1.5.2
helm install yunikorn yunikorn/yunikorn --namespace yunikorn \
  --set version=${VERSION}
```

If anyone agree with me, we can create a new Jira for the improvement.

chenyulin0719 (approved):
LGTM. Since this PR already addresses the request, I will merge it later today.

blueBlue0102:
> Sorry for the late review. Instead of providing a custom.yml, I would prefer to use a single command. ([Refer the chart configuration list](https://github.com/apache/yunikorn-release/tree/master/helm-charts/yunikorn#configuration))
> ...
> ...
> The even better solution is to extract variable `version` in helm's values.yml, so we could use:
> 
> ```
> VERSION=1.5.2
> helm install yunikorn yunikorn/yunikorn --namespace yunikorn \
>   --set version=${VERSION}
> ```
> 
> If anyone agree with me, we can create a new Jira for the improvement.

I agree with @chenyulin0719; using a single command is more convenient than providing a custom.yml.

chia7712:
> using a single command is more convenient than providing a custom.yml.

sounds good to me

chenyulin0719:
Created the followup Jira: [[YUNIKORN-2819] Add version parameter to Helm chart](https://issues.apache.org/jira/browse/YUNIKORN-2819)

PR #455, opened 2024-07-13
[YUNIKORN-2746] Adopting prometheus service monitor instead of modifying config

SP12893678:
### What is this PR for?
Adopting prometheus service monitor instead of modifying config

### What type of PR is it?
* [x] - Documentation

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2746


0yukali0 (approved):
+1 LGTM
I ran the ci and Let's see the result.

PR #486, opened 2024-10-10
[YUNIKORN-2914] Update deployment documentation for extra description…

chenyao0910:
### What is this PR for?

I just saw there's a user asking about configmap hot reload question in the slack channel, can refer to[https://yunikornworkspace.slack.com/archives/CLNUW68MU/p1728327140557209]

I also checked the hot refresh part for Yunikorn's docs, and I think not everyone knows that if you mounted the configmap under the subpath, the hot reload will not be triggered.

Therefore, I want to add extra description about this part. 

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2914`

### Questions:

### DIFF
--- a/docs/developer_guide/deployment.md
+++ b/docs/developer_guide/deployment.md
@@ -144,5 +144,8 @@ Once this is done, web UI will be available at: http://localhost:9889.
 
 YuniKorn uses config maps for configurations, and it supports loading configuration changes automatically by watching config map changes using shared informers.
 
-To make configuration changes, simply update the content in the configmap, which can be done either via Kubernetes dashboard UI or command line. Note, changes made to the configmap might have some delay to be picked up by the scheduler.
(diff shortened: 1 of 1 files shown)


craigcondit (changes requested):
The old information is accurate. The new information is not. We never mount the configmap at all. We use Kubernetes informers in the code to watch for changes automatically.

chenyao0910:
I get it. thx for your explanation.

I'll close the PR.

chenyao0910:
Another question, why we choose using informer rather than the hot reload by mount?

craigcondit:
As the scheduler, we use informers to track changes to many other K8s objects already. Extending that to the configmap was easy. It also means we are more responsive to changes (mounted configmaps can take up to 30 seconds to be updated).

PR #494, opened 2024-10-21
[YUNIKORN-2941] Remove plugin mode from the install section of Getting Started

kousei47747:
### What is this PR for?
Since plugin mode is now deprecated and will be removed in a future release, it would be better to remove the plugin mode in the install section to prevent any confusion.

### What type of PR is it?
* [ ] - Bug Fix
* [x] - Improvement
* [ ] - Feature
* [ ] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2941


ryankert01:
Maybe we should add a deprecated tag to it rather than delete it.

kousei47747:
@ryankert01 thanks for reviewing! 
I've updated it to add a deprecated tag. Please check.

craigcondit (changes requested):
Please remove the versioned docs update (we never update these in order to preserve historical context). I'd also prefer to see the entire deployment modes section removed from the getting started guide as it just confuses things now that we're getting rid of plugin mode.

kousei47747:
@craigcondit thanks for reviewing!
The versioned docs update and the entire deployment modes section in Getting Started have been removed, please check.

craigcondit (approved):
+1 LGTM.

PR #509, opened 2024-12-23
[YUNIKORN-2634] Remove deprecated usergroup resolution handling via the label

ryankert01:
### What is this PR for?
as discussed at [YUNIKORN-2207](https://issues.apache.org/jira/browse/YUNIKORN-2207).
should be done after 1.6 release.

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/browse/YUNIKORN-2634

### Questions:


wilfred-s (dismissed):
LGTM +1

chenyulin0719:
@wilfred-s , @craigcondit We didn't make deprecation announcement for `yunikorn.apache.org/username`  in 1.6.0 . Should we postpone this change to next release?

- https://yunikorn.apache.org/release-announce/1.6.0/

ryankert01:
Hi, @chenyulin0719. Maybe I should mark it as deprecated for now(rather than legacy)?

chenyulin0719:
I think we haven't discussed whether to deprecate the legacy label `yunikorn.apache.org/username`.
For back compatibility, I prefer keep it as legacy. If we are going to deprecate the label. I think a notification to user channel is required.

Need more feedback from other users/contributors.

craigcondit:
> I think we haven't discussed whether to deprecate the legacy label `yunikorn.apache.org/username`. For back compatibility, I prefer keep it as legacy. If we are going to deprecate the legacy label. I think a notification to user channel is required.

We should certainly deprecate it. We should probably document that as part of the 1.7.0 release notes and remove in 1.8.0.

wilfred-s:
There is more that needs to change on the page. We also need to log deprecation in 1.9 That changeis needed for the 1.9.0 release so we can remove the code in the upcoming 1.10 release.

PR #513, opened 2025-02-02
[Yunicorn-2773]  Removal of default queue configuration

kaichiachen:
### What is this PR for?
According to https://issues.apache.org/jira/browse/YUNIKORN-2711 and https://issues.apache.org/jira/browse/YUNIKORN-2703, user cannot config default queue name through admissionController.filtering.defaultQueue or other configuration anymore. Apps will be placed in 'root.default' if no rule is provided. 

### What type of PR is it?
* [ ] - Bug Fix
* [ ] - Improvement
* [ ] - Feature
* [x] - Documentation
* [ ] - Hot Fix
* [ ] - Refactoring

### Todos
* [ ] - Task

### What is the Jira issue?
https://issues.apache.org/jira/projects/YUNIKORN/issues/YUNIKORN-2773

Move this to the ‘Deprecated Settings’ section and add a link directing users to the ‘App Placement Rules’ page for guidance on configuring the desired queue name with placement rules.
**new added Deprecated Settings**
Screenshot 68
**Fixed Rule**
image

### Questions:

### DIFF
--- a/docs/user_guide/service_config.md
+++ b/docs/user_guide/service_config.md
@@ -1178,7 +1178,7 @@ service.operatorPlugins: "general"
   
 #### admissionController.filtering.defaultQueue
 
-**_DEPRECATED in 1.6.0:_** No replacement
+**_DEPRECATED in 1.6.0:_** This setting is deprecated. To configure a desire queue name, use the [Fixed Rule](placement_rules/#fixed-rule) placement rule instead.
 
 Controlls what will be the default queue name for the application.
 
--- a/docs/user_guide/service_config.md
+++ b/docs/user_guide/service_config.md
@@ -552,7 +552,6 @@ data:
   admissionController.filtering.labelNamespaces: ""
   admissionController.filtering.noLabelNamespaces: ""
   admissionController.filtering.generateUniqueAppId: "false"
-  admissionController.filtering.defaultQueue: "root.default"
   admissionController.accessControl.bypassAuth: "false"
   admissionController.accessControl.trustControllers: "true"
   admissionController.accessControl.systemUsers: "^system:serviceaccount:kube-system:"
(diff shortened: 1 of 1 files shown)


chenyulin0719 (commented):
Hi @kaichiachen, thanks for taking over the Jira.

Although we removed the default queue configuration, we can still achieve the same behavior by setting the `fixed` rule in the end of placement rules. I'm in favor of adding such an example under the deprecated config. 

For example, this explanation looks pretty good to me.
- https://github.com/apache/yunikorn-k8shim/pull/874#issuecomment-2216653252

@kaichiachen WDYT?

kaichiachen:
> Hi @kaichiachen, thanks for taking over the Jira.
> 
> Although we removed the default queue configuration, we can still achieve the same behavior by setting the `fixed` rule in the end of placement rules. I'm in favor of adding such an example under the deprecated config.
> 
> For example, this explanation looks pretty good to me.
> 
> * [[YUNIKORN-2711] Skip setting the queue name to default queue in the shim yunikorn-k8shim#874 (comment)](https://github.com/apache/yunikorn-k8shim/pull/874#issuecomment-2216653252)
> 
> @kaichiachen WDYT?

Thanks @chenyulin0719 ! I agree that we should provide guidance on configuring the desired queue once this configuration is no longer available. The official documentation includes a section elaborating how to configure the desired queue name: https://yunikorn.apache.org/docs/next/user_guide/placement_rules#fixed-rule 
I think we can leverage it and add a hyperlink to the deprecated configuration. You can check my latest screenshot, thanks!

chenyulin0719 (approved):
+1 LGTM, adding hyperlink make sense to me.

docs/user_guide/service_config.md
```diff
@@ -1204,3 +1175,34 @@ service.operatorPlugins: "general"
     - shim.scheduler
     - shim.scheduler.plugin
     - shim.utils
+  
+#### admissionController.filtering.defaultQueue
```
manirajv06, line 1179: Should we have this in Deprecated settings section, that too after "service.operatorPlugins" section?

manirajv06 (changes requested):

docs/user_guide/service_config.md
kaichiachen: Yes, I think this should be in the Deprecated Settings section because yunikorn won't provide config "admissionController.filtering.defaultQueue" anymore, and I have already made the change after the service.operatorPlugins section.

chenyulin0719: @kaichiachen I guess what @manirajv06 means is the order of `Footnotes` and `deprecated setting` in markdown file. Although markdown automatically move footnotes to the end of the render page, it's still strange to see them in service_config.md file.

```
### Deprecated settings

#### service.operatorPlugins
**_DEPRECATED in 1.4.0:_** No replacement
...ignored

[^1]: Available log subsystem values:        -> I think it should be moved to the end of md file,  or placed right after the `log.{subsystem}.level`.
    - admission
    - admission.client
    - admission.conf
...ignored

#### admissionController.filtering.defaultQueue

**_DEPRECATED in 1.6.0:_** 
```

chenyulin0719 (changes requested):
@kaichiachen The result layout is correct. This is a very nitpicky change request.

docs/user_guide/service_config.md
kaichiachen: Good catch! That's a fair point. Thanks @chenyulin0719 for clarification!

chenyulin0719 (approved):
LGTM

PR #557, opened 2026-06-09
[YUNIKORN-3297] Update license check

wilfred-s:
### What is this PR for?
Remove pnpm-workspace.yaml from the files checked. Update regexp to ignore ./build and all the artifacts below it as it is the website production build which does not have license headers in the js files.

### What type of PR is it?
* [X] - Bug Fix

### What is the Jira issue?
* Jira issue: [YUNIKORN-3297](https://issues.apache.org/jira/browse/YUNIKORN-3297)

./check_license.sh should pass in CI and local builds even after a doing a production build on the local machine


adityadtu5 (approved):

manirajv06 (approved):
looks good

PR #569, opened 2026-08-03
[YUNIKORN-3266] Add remote container debugging guide to developer docs

kaijaytu:
### Description
Add a new "Debug scheduler in a remote cluster" section to the developer
guide (env_setup.md). The existing documentation covers running the scheduler
locally with `KUBECONFIG` and switching kubectl context to a remote cluster,
but does not explain how to deploy a debug-enabled image and attach a remote
debugger.

This section covers:
- Prerequisites (remote cluster access, container registry, Delve)
- Building a debug image with `-gcflags="all=-N -l"` and a multi-stage Dockerfile
- Pushing the image to a container registry
- Deploying to a remote cluster with the debug port exposed
- Attaching a debugger from GoLand or VS Code via port-forwarding
- Cleanup notes (do not use debug images in production)

### Type of change
- [x] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3266

### AI Tooling
- [x] The PR includes the phrase "Generated by GitHub Copilot (Claude Opus 4.6)"

### How Has This Been Tested?
- [x] `pnpm build` generated static files successfully

Documentation-only change. No functional code modified.

### DIFF
--- a/docs/developer_guide/env_setup.md
+++ b/docs/developer_guide/env_setup.md
@@ -255,3 +255,108 @@ Or follow these simplified steps:
 
 More documentation can be found
 [here](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/).
+
+## Debug scheduler in a remote cluster
+
+This section explains how to deploy a debug-enabled scheduler image to a remote cluster and attach a debugger from your local machine using [Delve](https://github.com/go-delve/delve).
+
+### Prerequisites
+
+- A remote Kubernetes cluster accessible via `kubectl` (see [above](#access-remote-kubernetes-cluster))
+- A container registry you can push to (Docker Hub, private registry, etc.)
+- [Delve](https://github.com/go-delve/delve) installed locally:
+  ```shell script
+  go install github.com/go-delve/delve/cmd/dlv@latest
+  ```
+
+### Build a debug image
+
+Compile the scheduler with optimizations and inlining disabled so that the debugger can map breakpoints correctly:
+
+```shell script
+cd yunikorn-k8shim
+CGO_ENABLED=0 go build -gcflags="all=-N -l" -o _output/shim ./pkg/cmd/shim
+```
+
+Create a Dockerfile for the debug image. This installs Delve and uses it as the entrypoint:
+
+```dockerfile
+FROM golang:1.22 AS dlv-builder
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
+FROM ubuntu:22.04
+COPY --from=dlv-builder /go/bin/dlv /usr/local/bin/dlv
+COPY _output/shim /scheduler
+EXPOSE 2345
+ENTRYPOINT ["dlv", "exec", "/scheduler", "--headless", "--listen=:2345", "--api-version=2", "--accept-multiclient"]
+```
+
+Build and tag the image:
+
+```shell script
+docker build -t <your-registry>/yunikorn:scheduler-debug -f Dockerfile.debug .
+```
+
+### Push image to registry
+
+```shell script
+docker push <your-registry>/yunikorn:scheduler-debug
+```
+
+Alternatively, update the `REGISTRY` variable in the `Makefile` (see [Build Docker images](build.md#build-docker-images)) and adapt the build target.
+
+### Deploy to remote cluster
+
+Update the scheduler Deployment (or Helm values) to use the debug image and expose port 2345:
+
+```yaml
+spec:
+  containers:
+    - name: yunikorn-scheduler
+      image: <your-registry>/yunikorn:scheduler-debug
+      ports:
+        - containerPort: 2345
+          name: dlv
+          protocol: TCP
+```
+
+Deploy using Helm or `kubectl apply`. Ensure the Pod has sufficient permissions (the same ServiceAccount used for normal scheduler deployment).
+
+### Attach debugger from local machine
+
+Port-forward the debug port:
+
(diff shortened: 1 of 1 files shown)


manirajv06 (approved):
looks good

manirajv06:
@adityadtu5 Can you take a look?

docs/developer_guide/env_setup.md
```diff
@@ -255,3 +255,108 @@ Or follow these simplified steps:
...
+  go install github.com/go-delve/delve/cmd/dlv@latest
+  ```
+
+### Build a debug image
+
+Compile the scheduler with optimizations and inlining disabled so that the debugger can map breakpoints correctly:
+
+```shell script
+cd yunikorn-k8shim
+CGO_ENABLED=0 go build -gcflags="all=-N -l" -o _output/shim ./pkg/cmd/shim
```
adityadtu5, line 278: nit: please also document what each flag does

docs/developer_guide/env_setup.md
```diff
@@ -255,3 +255,108 @@ Or follow these simplified steps:
...
+docker build -t <your-registry>/yunikorn:scheduler-debug -f Dockerfile.debug .
+```
+
+### Push image to registry
+
+```shell script
+docker push <your-registry>/yunikorn:scheduler-debug
+```
+
+Alternatively, update the `REGISTRY` variable in the `Makefile` (see [Build Docker images](build.md#build-docker-images)) and adapt the build target.
```
adityadtu5: Since this is not a makefile target, updating `REGISTRY` variable there won't help

adityadtu5 (commented):
Mostly looks good, some minor comments

docs/developer_guide/env_setup.md
```diff
@@ -255,3 +255,108 @@ Or follow these simplified steps:
...
+
+```shell script
+cd yunikorn-k8shim
+CGO_ENABLED=0 go build -gcflags="all=-N -l" -o _output/shim ./pkg/cmd/shim
+```
+
+Create a Dockerfile for the debug image. This installs Delve and uses it as the entrypoint:
+
+```dockerfile
+FROM golang:1.22 AS dlv-builder
```
wilfred-s: nit: can be fixed on commit we should use go 1.26 at least, it is the go repro version and what we build with for master at the moment.

docs/developer_guide/env_setup.md
wilfred-s: We can use `REGISTRY` shell variable in the call to make: `REGISTRY="reg.local.host" make image`

Additionally we can add building debug images into the k8shim makefile like we have done for reproducible builds: `REPRODUCIBLE_BUILDS=1 make image` make it something like `DEBUG_BUILDS=1 make image`
based on the flag we add `-gcflags="all=-N -l"` to the build command. 
Adding a variable in the command line we build, example after the `-a` option
```
	-a ${GCFLAGS} \
```
If DEBUG_BUILDS is not set GCFLAGS is empty otherwise it is set to `GCFLAGS='-gcflags="all=-N -l"'`
It can then be combined with the `REGISTRY` variable and keep everything in one place.

wilfred-s: We can then even override the VERSION inside the Makefile from `latest` to `debug` and have it properly tracked without overriding production images...
That can be extended as far as we want/need...

docs/developer_guide/env_setup.md
kaijaytu: Good catch, I will add a brief explanation for each flag:

- CGO_ENABLED=0: produce a statically linked binary so it can run inside the minimal debug image without extra libc dependencies.
- -gcflags="all=-N -l": applied to all packages; -N disables optimizations and -l disables inlining, so Delve can map source lines to instructions and breakpoints hit reliably.
- -o _output/shim: output path that matches the COPY _output/shim /scheduler line in the Dockerfile below.

Will push the update together with the other review fixes.

docs/developer_guide/env_setup.md
kaijaytu: You are right, that sentence is misleading. There is no dedicated debug-image target in the k8shim Makefile today, so editing REGISTRY in the Makefile alone does not help build or push the debug image.

I will drop that line and rewrite it around the docker tag / docker push flow the user has to run manually.

As a side note: @wilfred-s left a follow-up suggesting a cleaner path -- adding DEBUG_BUILDS=1 make image support in the k8shim Makefile so users can do REGISTRY=<reg> DEBUG_BUILDS=1 make image instead of the manual go build + custom Dockerfile shown here. That is a bigger change (touches k8shim, not just the docs), so I will handle the wording fix in this PR first and open a separate discussion for the Makefile enhancement. Let me know if you would prefer the doc to already reference the future flag.

docs/developer_guide/env_setup.md
kaijaytu: Thanks, will bump golang:1.22 to golang:1.26 (the current master repro version) in the next push, along with the other review fixes.

wilfred-s (approved):
LGTM

docs/developer_guide/env_setup.md
wilfred-s: Please file the followup jira so we do not forget @kaijaytu

kaijaytu: Filed follow-up as YUNIKORN-3407:
https://issues.apache.org/jira/browse/YUNIKORN-3407
Will pick it up next.

PR #565, opened 2026-07-07
[YUNIKORN-3247] Update troubleshooting statedump collection

PoiBlackTea:
Update the troubleshooting documentation for the collection of the statedump. Replace the deprecated /ws/v1/fullstatedump endpoint and port 9889 with the correct /debug/fullstatedump endpoint and port 9080, which is hosted by the core scheduler web service directly. Add port-forwarding instructions for port 9080.

### Description
Short description of this pull request, can be as simple as the commit message used.
First time contributing? Check out the contributing guide: https://yunikorn.apache.org/community/how_to_contribute

### Type of change
Please delete options that are not relevant.

- [X] Documentation
- [ ] Improvement
- [ ] Feature
- [ ] Bug Fix

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3247

- [X] I have created a Jira issue for this pull request
- [X] The Jira ID is part of the title of this pull request

### AI Tooling
If an AI tool was used:
- [X] The PR includes the phrase "Generated by Antigravity IDE", where Antigravity IDE is the name of the AI tool used.
- [X] My use of AI contributions follows the ASF legal policy.

Check https://www.apache.org/legal/generative-tooling.html for details.

### How Has This Been Tested?
- [X] `check_license.sh` returned an "all OK" result
- [X] `local_build.sh run` generated a usable website

### Screenshots or other details
NA

PR #564, opened 2026-07-02
[YUNIKORN-3310] Document yunikorn_queue_app_total metric

urguru:
### Description
Add documentation for the new `yunikorn_queue_app_total` counter metric that tracks cumulative terminal application states (completed, failed, rejected) per queue.

This is the documentation follow-up for the code change merged in apache/yunikorn-core#1097.

### Type of change
- [x] Documentation

image

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3310

- [x] I have created a Jira issue for this pull request.
- [x] The Jira ID is part of the title of this pull request.

### AI Tooling
- [x] The PR includes the phrase "Generated by Claude Code", where Claude Code is the name of the AI tool used.
- [x] My use of AI contributions follows the ASF legal policy.

Generated by Claude Code

PR #563, opened 2026-06-30
[YUNIKORN-3314] Update logging subsystem values

wilfred-s:
### Description
Update the supported log subsystem values in the documentation. Added:
* deprecation (now supported in K8shim also)
* core.diagnostics (1.5 ?)
* core.scheduler.nodesusage (1.5)
* core.scheduler.preemption.quotachange (1.8)
* core.scheduler.preemption.requirednode (1.8)
* shim.placeholder.config (1.9)

Removed:
* shim.appmgmt (1.6 ?)
* shim.appmgmt.general (1.6 ?)
* shim.appmgmt.sparkoperator (1.7)
* shim.scheduler.plugin (1.9)

### Type of change
- [X] Documentation

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3314

- [X] I have created a Jira issue for this pull request
- [X] The Jira ID is part of the title of this pull request

### AI Tooling
If an AI tool was used:
- [ ] The PR includes the phrase "Generated by \<tool>", where \<tool> is the name of the AI tool used.
- [ ] My use of AI contributions follows the ASF legal policy.

Check https://www.apache.org/legal/generative-tooling.html for details.

### How Has This Been Tested?
- [X] `check_license.sh` returned an "all OK" result
- [X] `local_build.sh run` generated a usable website

PR #571, opened 2026-08-07
[YUNIKORN-3348] Fix and extend the concurrency documentation

tigerquoll:
### Description

The only statement of the scheduler's lock ordering rule anywhere on the site was malformed, and the text around it predates the deadlock detection feature added in YUNIKORN-2539.

- **`docs/design/cache_removal.md`** — correct the lock direction rule. The second bullet read *"Holding B.lock and acquire B's lock"*, which is a tautology, so the rule as published stated nothing. It should be *"acquire A's lock"*. This is the only lock-ordering statement in the documentation.
- **`docs/design/cache_removal.md`** — replace *"There are no known tools that could be used to detect or describe lock order"* with a reference to `pkg/locking`. That has been inaccurate since YUNIKORN-2539 (commit `5758d7a`, 2024-04-05) added the package, which wraps [go-deadlock](https://github.com/sasha-s/go-deadlock) and performs exactly lock order and lock wait detection.
- **`docs/user_guide/troubleshooting.md`** — new "Deadlock detection" section documenting the four `DEADLOCK_*` environment variables, the two checks they control, where the report is logged (`core.diagnostics`, `ERROR`), and how to read the current setting back from `/ws/v1/config`.
- **`docs/developer_guide/build.md`** — state that `make test` runs with `-race` and deadlock detection enabled, so a run ending in `POTENTIAL DEADLOCK` is a real lock problem rather than a flaky test.

#### Two notes for reviewers

**Placement.** The Jira suggests `service_config.md` for the environment variables. I have put them in `troubleshooting.md` instead: `service_config.md` documents Helm values and ConfigMap keys, and these are neither — they are environment variables set on the container. `troubleshooting.md` already covers log retrieval, the state dump and restarting the scheduler, which is the context an operator is in when they need this. Happy to move it if you would prefer to keep all runtime settings together.

**Versioned docs.** This changes `docs/` only, matching recent documentation PRs (#563, #564, #565), so the corrections appear in the next release. The published 1.9.0 docs keep the garbled rule until then. Let me know if you would like the two `cache_removal.md` corrections backported to `versioned_docs/version-1.9.0/` as well — they are factual errors rather than new content, so there is an argument for it.

### Type of change

- [x] Documentation
- [x] Bug Fix

### Jira issue

Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3348

### AI Tooling

Generated by the Author with assistance from Claude Code.

- [x] The PR includes the phrase "Generated by \<tool>", where \<tool> is the name of the AI tool used.

### How Has This Been Tested?

- [ ] `local_build.sh run` generated a usable website

Instead of the Docker wrapper, the site was built directly with `pnpm install --frozen-lockfile && pnpm build`, which runs the same Docusaurus build. Result: `[SUCCESS] Generated static files in "build"`.

Checks performed on the generated output:

- The new `## Deadlock detection` heading produces the anchor `#deadlock-detection` in `build/docs/next/user_guide/troubleshooting/index.html`.
- Both new cross-document links resolve, appearing in the generated HTML as `href="/docs/next/user_guide/troubleshooting#deadlock-detection"` from `design/cache_removal` and `developer_guide/build`.
- The build reports pre-existing broken-anchor warnings elsewhere on the site; none of them are on the three pages touched here, and this change adds none.

Sources for the factual claims in the new documentation, so they can be checked:

- The four environment variables, their defaults and behaviour: `yunikorn-core/pkg/locking/locking.go`.
- The shim delegates its configuration to the core, so one set of settings covers both: `yunikorn-k8shim/pkg/locking/locking.go` calls into `corelocking` from its `init()`.
- Reports are logged under `core.diagnostics` at `ERROR` level: `printBufContents()` in the same core file.
- `/ws/v1/config` returns `DeadlockDetectionEnabled` and `DeadlockTimeoutSeconds`: `yunikorn-core/pkg/webservice/dao/config_info.go` and `handlers.go`.
- `make test` exports `DEADLOCK_DETECTION_ENABLED=true`, `DEADLOCK_TIMEOUT_SECONDS=10`, `DEADLOCK_EXIT=true` and runs with `-race` in both repositories: the `test` target in each `Makefile`.

### DIFF
--- a/docs/design/cache_removal.md
+++ b/docs/design/cache_removal.md
@@ -423,7 +423,7 @@ A state change is thus immediate and this should prevent an issue like [YUNIKORN
 ### Direction of lock 
 It is possible to acquire another lock while holding a lock, but we need to make sure that we do not allow: 
 - Holding A.lock and acquire B's lock. 
-- Holding B.lock and acquire B's lock. 
+- Holding B.lock and acquire A's lock. 
 
 The current code in the scheduler takes a lock as late as possible and only for the time period needed.
 Some actions are not locked on the scheduler side just on the cache side as each object has its own lock.
@@ -453,4 +453,8 @@ The partition should be locked while retrieving for instance the node that needs
 
 This approach fits in with the current locking approach and will keep the locking changes to a minimum.
 Testing, specifically end-to-end testing, should catch these deadlocks. 
-There are no known tools that could be used to detect or describe lock order.
+
+Since this design was written, runtime detection has been added.
+The `pkg/locking` package wraps [go-deadlock](https://github.com/sasha-s/go-deadlock) and provides both lock order detection and lock wait timeout detection for all locks that use it.
+It is disabled by default and is turned on using the environment variables described in [deadlock detection](../user_guide/troubleshooting.md#deadlock-detection).
+The unit test run, `make test`, enables it in both the core and the k8shim.
--- a/docs/developer_guide/build.md
+++ b/docs/developer_guide/build.md
@@ -163,6 +163,12 @@ To make sure that the local changes will not break other parts of the
 build you should run:
 - A full build `make` (build target depends on the repository)
 - A full unit test run `make test`
+  This runs the tests with the race detector (`-race`) and with deadlock detection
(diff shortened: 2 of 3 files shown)


manirajv06 (approved):
looks good

PR #574, opened 2026-08-23
[YUNIKORN-3400] Remove outdated quota preemption inheritance note

andyhuangdev:
### Description

Remove the outdated note stating that `quota.preemption.delay` inheritance is unsupported.

Queue properties are already inherited during child queue creation, and the behavior was verified by YUNIKORN-3208. The note is removed from both the current documentation and the latest published 1.9.0 documentation.

### Type of change

- [x] Documentation

### Jira issue

Jira ID: https://issues.apache.org/jira/browse/YUNIKORN-3400

### How Has This Been Tested?

- [x] The Docusaurus production build completed successfully
- [x] Verified that the current and 1.9.0 quota preemption pages no longer contain the outdated inheritance note


wilfred-s (commented):
LGTM

Note that we normally do not update already published doc versions but in this case the jira explicitly mentioned is marked as fixed in the release.

PR #570, opened 2026-08-03
[YUNIKORN-3248] Improve quota preemption documentation

kaijaytu:
### Description
Improve the quota preemption user guide with missing details identified in YUNIKORN-3248:
- Clarify that `quota.preemption.delay` uses Go `time.Duration` format (e.g. `15s`, `5m`, `1h30m`)
- Add "Triggering preconditions" section listing all 6 conditions required for quota preemption
- Add "Configuration change scenarios" table covering max decrease, delay change, consecutive decreases, quota increase back, and max removal
- Add "Restart behavior" section explaining that the preemption timer is in-memory only and resets on restart

**Note on pending pods (JIRA item #3):** The JIRA suggests documenting that quota preemption only triggers when pending pods exist. After reviewing the code (`tryAcquirePreemption()` in `queue.go`, `internalQuotaPreemption()` in `scheduler.go`, and `quota_preemptor.go`), there is no pending pods check in the quota preemption flow. It triggers solely based on `allocatedResource > maxResource`. This is documented as a note in the "Triggering preconditions" section.

**Observation:** `pnpm build` reports 50 pre-existing broken anchors across versioned docs (mostly 1.0.01.3.0), unrelated to this change. Would it be worth filing a separate issue to clean those up?

Generated by GitHub Copilot (Claude Opus 4.6)

### Type of change
- [x] Documentation
- [x] Improvement

### Jira issue
Jira ID : https://issues.apache.org/jira/browse/YUNIKORN-3248

- [x] I have created a Jira issue for this pull request
- [x] The Jira ID is part of the title of this pull request

### AI Tooling
- [x] The PR includes the phrase "Generated by GitHub Copilot (Claude Opus 4.6)"
- [x] My use of AI contributions follows the ASF legal policy.

### How Has This Been Tested?
- [x] `check_license.sh` returned an "all OK" result
- [x] `local_build.sh run` generated a usable website

### Screenshots or other details
N/A

PR #575, opened 2026-08-26
[YUNIKORN-3388] Update quota_preemptor design doc to match current implementation

kaijaytu:
JIRA: https://issues.apache.org/jira/browse/YUNIKORN-3388

Three items in docs/design/quota_preemptor.md were out of sync with
the current implementation in apache/yunikorn-core:

1. Delay format and placement
   Doc showed `quota.preemption.delay` as an integer-seconds field
   under queue `resources:`. Implementation reads it from queue
   `properties:` and parses it with `time.ParseDuration`. Updated
   both the generic YAML block and the two examples to use
   `properties: quota.preemption.delay: <duration>` with values
   like `2h` / `5s`.

2. Timer behavior on config changes
   Doc claimed "timers reset and start again from the beginning".
   Actual behavior in `setPreemptionTime()`:
   - Delay change (with or without max change): existing start time
     is shifted by (newDelay - oldDelay).
   - Consecutive quota decreases: the earliest start time is
     preserved (only the delta adjustment above is applied if the
     delay also changed).
   - Delay was 0 and is now positive with no max change: a fresh
     `now + newDelay` start time is set.
   Reworded the paragraph accordingly.

3. Impact of Restart
   Added the fact that during recovery `IncAllocatedResource()`
   sets `quotaPreemptionStartTime = now + quotaPreemptionDelay`
   when usage exceeds max, which is what effectively restarts the
   countdown from the recovery moment.

Also fixed a stale internal link in the "Preemption Fence &
Preemption Delay" section that pointed to
`Queue=>resources=>preemption.delay(#quota-preemption-configuration)`;
it now points to
`Queue=>properties=>quota.preemption.delay(#quota-preemption-delay)`.

Verified: `pnpm build` runs to completion; no new broken-anchor
warnings tied to the edited file. Reverse-link scan clean.


kaijaytu:
@manirajv06 @wilfreds could you take a look when you have time?
This is the design-doc follow-up to YUNIKORN-3248 (#570).
Thanks!

HuangTing-Yao (approved):
LGTM — verified against `yunikorn-core@master`: the `properties: quota.preemption.delay` duration format, timer-adjustment, and restart behavior all match the implementation (`setPreemptionTime` / `IncAllocatedResource`).

